PR #4, opened 2018-09-12
Drop Python 3.5 support

cburgdorf:
## What was wrong?

Issue #

## How was it fixed?

Summary of approach.

#### Cute Animal Picture


pipermerriam:
Need to add 3.7 too (though not necessarily here)

Issue #9, opened 2018-09-20
Several APIs (including stream) should accept a CancelToken parameter


cburgdorf:
## What was wrong?

There's currently no builtin way to unsubscribe from the `stream` API. See discussion: https://github.com/ethereum/py-evm/pull/1310#discussion_r219179856

## How can it be fixed?

e.g.
`stream(SomeEvent, cancel_token=token)`

pipermerriam:
Closing as I don't think we're doing this anymore.  See this comment for why: https://github.com/ethereum/lahja/pull/11#pullrequestreview-158660912

PR #11, opened 2018-09-23
Make subscribe accept and return CancelToken

cburgdorf:
## What was wrong?

As described in #9 we need to add native support for `CancelToken`

## How was it fixed?

Added first class `CancelToken` support for all APIs

#### Cute Animal Picture

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,22 +124,22 @@ Events can be received in three different fashions. All APIs are non-blocking.
 
 **Subscribe API**
 
-`def subscribe(self, event_type: Type[BaseEvent], handler: Callable[[BaseEvent], None]) -> Subscription:`
+`def subscribe(self, event_type: Type[BaseEvent], handler: Callable[[BaseEvent], None]) -> CancelToken:`
 
 *Example:*
 
 ```Python
-subscription = endpoint.subscribe(SecondThingHappened, lambda event: 
+cancel_token = endpoint.subscribe(SecondThingHappened, lambda event: 
     print("Received via SUBSCRIBE API in proc1: ", event.payload)
 )
 ```
 
-The handler will be called every time that a `SecondThingHappened` event is fired. Notice that the returned `Subscription` allows deregistering from the event at any later point in time.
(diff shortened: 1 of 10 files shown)


cburgdorf:
Here's something I've thought about. When we talked about that the `CancelToken` is really only useful for `stream` as something like `request` could be wrapped in a `wait` I think that this may only be half of the truth.

While it is true that the wrapping something like `request` in a `wait` will ensure that we can stop awaiting the request, I'm not sure if this will allow us to do the proper internal cleanup of the underlying registered future.

E.g. here is how `request` is implemented using `CancelToken`.

```python

async def request(self,
                      item: BaseRequestResponseEvent[TResponse],
                      cancel_token: Optional[CancelToken] = None) -> TResponse:
        item._origin = self.name
        item._id = str(uuid.uuid4())

        future: asyncio.Future = asyncio.Future()
        self._futures[item._id] = future

        self._sending_queue.put_nowait((item, None))

        token = self._chain_or_create_cancel_token(f'request#{item}', cancel_token)

        try:
            result = await token.cancellable_wait(future)
        except OperationCancelled as e:
            del self._futures[item._id]
            raise e

        expected_response_type = item.expected_response_type()
        if not isinstance(result, expected_response_type):
            raise UnexpectedResponse(
                f"The type of the response is {type(result)}, expected: {expected_response_type}"
            )

        return result
```

Notice how we are doing `del self._futures[item._id]` to cleanup the future. 

Now, I'm trying to figure out how we could deal with that if for instance the callsite does:

```python
await asyncio.wait_for(endpoint.request(item), 0.1)
```

I thought that maybe inside `request` I could implement it as such:

```python
        try:
            result = await future
        except asyncio.CancelledError as e:
            del self._futures[item._id]
            raise e
```

But that doesn't seem to do the trick so I keep digging.

Another idea, we could also consider to support `CancelToken` but use something like a *protocol* where we aren't depending on `ethereum/async-cancel-token` but rather try to make it work with any kind of `CancelToken` that supports that *protocol*. Lahja could then ship its very own minimalistic `CancelToken` implementation. Anyway I keep digging.

PR #12, opened 2018-09-25
Some minor fixes and cleanup

cburgdorf:
## What was wrong?

Since it seems we won't add `CancelToken` support as proposed in #9, here are a couple of things that I extracted from that PR:

- A fix were we tried to propagate events when the event bus was already shut down
- fixed broken example
- tightened tests and assertions
- basic cleanup

## How was it fixed?

:fire: 

#### Cute Animal Picture

### DIFF
--- a/tests/core/conftest.py
+++ b/tests/core/conftest.py
@@ -1,0 +1,24 @@
+import asyncio
+from typing import (
+    Generator,
+)
+
+import pytest
+
+from lahja import (
+    Endpoint,
+    EventBus,
+)
+
+
+@pytest.fixture(scope='function')
+def endpoint(event_loop: asyncio.AbstractEventLoop) -> Generator[Endpoint, None, None]:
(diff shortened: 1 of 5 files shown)


tests/core/conftest.py
```diff
@@ -0,0 +1,33 @@
...
+import pytest
+
+from lahja import (
+    Endpoint,
+    EventBus,
+)
+
+
+@pytest.fixture(scope='session')
+def event_loop() -> Generator[asyncio.AbstractEventLoop, None, None]:
```
pipermerriam: I was pretty sure that `pytest-asyncio` already exposes an `event_loop` fixture automatically.  If there is a reason we aren't using that one or why we're overriding it can you add a comment explaining.

cburgdorf: :+1: Didn't know that. Works! Will update and merge.

PR #13, opened 2018-09-26
Ensure cancelled `request` does not leave stale registration behind

cburgdorf:
## What was wrong?

Currently if the consumption of `bus.request(Something()` is cancelled, we are leaving a stale registration behind.

## How was it fixed?

Detect cancelled Future and remove registration.

#### Cute Animal Picture


cburgdorf:
@pipermerriam I believe this is the right pattern to remove cruft when the consumption of a `request(Something())` was cancelled.

PR #14, opened 2018-09-27
Ensure streams can get cancelled

cburgdorf:
## What was wrong?

It was unclear how streams could be cancelled.

## How was it fixed?

Added a fix + test to demo how streams can get cancelled

#### Cute Animal Picture

### DIFF
--- a/lahja/endpoint.py
+++ b/lahja/endpoint.py
@@ -21,6 +21,7 @@ from .async_util import (
     async_get,
 )
 from .exceptions import (
+    NoConnection,
     UnexpectedResponse,
 )
 from .misc import (
@@ -48,13 +49,22 @@ class Endpoint:
         self._running = False
         self._executor: Optional[ThreadPoolExecutor] = None
 
+    @property
+    def loop(self) -> asyncio.AbstractEventLoop:
+        if self._loop is None:
+            raise NoConnection("Need to call `connect()` first")
+        return self._loop
+
     def connect(self, loop: Optional[asyncio.AbstractEventLoop] = None) -> None:
         """
         Connect the :class:`~lahja.endpoint.Endpoint` to the :class:`~lahja.eventbus.EventBus`
         instance that created this endpoint.
         """
-        # mypy doesn't recognize loop as Optional[AbstractEventLoop].
-        asyncio.ensure_future(self._try_connect(loop), loop=loop)  # type: ignore
+        if (loop is None):
+            loop = asyncio.get_event_loop()
+
+        self._loop = loop
+        asyncio.ensure_future(self._try_connect(loop), loop=loop)
 
     async def _try_connect(self, loop: asyncio.AbstractEventLoop) -> None:
         # We need to handle exceptions here to not get `Task exception was never retrieved`
@@ -130,7 +140,7 @@ class Endpoint:
         item._origin = self.name
         item._id = str(uuid.uuid4())
 
-        future: asyncio.Future = asyncio.Future()
+        future: asyncio.Future = asyncio.Future(loop=self.loop)
         self._futures[item._id] = future
 
         self._sending_queue.put_nowait((item, None))
--- a/lahja/endpoint.py
+++ b/lahja/endpoint.py
@@ -191,16 +191,21 @@ class Endpoint:
         self._queues[event_type].append(queue)
         i = None if max is None else 0
         while True:
-            event = await queue.get()
-            if i is not None:
-                i += 1
             try:
-                yield event
+                yield await queue.get()
             except GeneratorExit as e:
                 self._queues[event_type].remove(queue)
                 raise e
+            except asyncio.CancelledError as e:
+                self._queues[event_type].remove(queue)
+                raise e
             else:
-                if i is not None and i >= cast(int, max):
+                if i is None:
+                    continue
+
+                i += 1
+
+                if i >= cast(int, max):
                     self._queues[event_type].remove(queue)
                     break
 
--- a/lahja/endpoint.py
+++ b/lahja/endpoint.py
@@ -186,11 +186,11 @@ class Endpoint:
 
     async def stream(self,
                      event_type: Type[TStreamEvent],
-                     max: Optional[int] = None) -> AsyncGenerator[TStreamEvent, None]:
+                     num_events: Optional[int] = None) -> AsyncGenerator[TStreamEvent, None]:
         """
         Stream all events that match the specified event type. This returns an
         ``AsyncIterable[BaseEvent]`` which can be consumed through an ``async for`` loop.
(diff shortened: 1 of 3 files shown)


tests/core/test_basics.py
```diff
@@ -139,6 +139,40 @@ def expected_response_type() -> Type[DummyResponse]:
...
+            print(event.property_of_dummy_request)
+            nonlocal stream_counter
+            stream_counter += 1
+            await asyncio.sleep(0.5)
+
+    async def cancel_soon() -> None:
+        while True:
+            await asyncio.sleep(0.01)
+            if stream_counter == 2:
+                await agen.aclose()
```
cburgdorf: @pipermerriam I did some research and I believe this is the default way to cancel an `AsyncGenerator`. See: https://www.python.org/dev/peps/pep-0525/#id14

tests/core/test_basics.py
```diff
@@ -139,6 +139,40 @@ def expected_response_type() -> Type[DummyResponse]:
     assert stream_counter == 2
 
 
+@pytest.mark.asyncio
+async def test_stream_can_get_cancelled(endpoint: Endpoint) -> None:
+    stream_counter = 0
+
+    async_generator = endpoint.stream(DummyRequest)
```
cburgdorf: If we want to be able to cancel our stream, we have to save a reference to the returned async generator (instead of immediately `async for`'ing over it). We can then later say `await async_generator.aclose()`

tests/core/test_basics.py
```diff
@@ -139,6 +139,40 @@ def expected_response_type() -> Type[DummyResponse]:
...
+            print(event.property_of_dummy_request)
+            nonlocal stream_counter
+            stream_counter += 1
+            await asyncio.sleep(0.5)
+
+    async def cancel_soon() -> None:
+        while True:
+            await asyncio.sleep(0.01)
+            if stream_counter == 2:
+                await async_generator.aclose()
```
cburgdorf: @pipermerriam I did some research and I believe this is the default way to cancel an `AsyncGenerator`. See: https://www.python.org/dev/peps/pep-0525/#id14

cburgdorf: Related: It looks like we should change out tear down code everywhere were event loops are used to look like this:

```python
try:
    loop.run_forever()
finally:
    loop.run_until_complete(loop.shutdown_asyncgens())
    loop.close()
```
See: https://docs.python.org/3/library/asyncio-eventloop.html?highlight=loop#asyncio.loop.shutdown_asyncgens

tests/core/test_basics.py
```diff
@@ -190,3 +190,13 @@ def expected_response_type() -> Type[DummyResponse]:
 
     await asyncio.sleep(0.01)
     assert isinstance(received, DummyRequest)
+
+
+@pytest.mark.asyncio
+async def test_wait_for_can_get_cancelled(endpoint: Endpoint) -> None:
+
+    with pytest.raises(asyncio.TimeoutError):
+        await asyncio.wait_for(endpoint.wait_for(DummyRequest), 0.1)
```
cburgdorf: On a different note: I think `wait_for` is a poor API name. I'm usually bad with naming things but here is some inspiration:

- `wait_for_event`
- `once`
- `single`
- `single_event`

pipermerriam: Finding myself *mildly* resistant to adding the `_event` suffix since it feels redundant (since the argument is an event).  No strong feelings.

lahja/endpoint.py
```diff
@@ -191,15 +201,21 @@ def subscribe(self,
...
                 self._queues[event_type].remove(queue)
+                raise e
             else:
-                if i is not None and i >= cast(int, max):
+                if i is None:
+                    continue
+
+                i += 1
+
+                if i >= cast(int, max):
```
pipermerriam: we should use something other than `max` to avoid overwriting the builtin `max` function.  Maybe `num_events`?

tests/core/test_basics.py
```diff
@@ -139,6 +139,40 @@ def expected_response_type() -> Type[DummyResponse]:
...
+    stream_counter = 0
+
+    async_generator = endpoint.stream(DummyRequest)
+
+    async def stream_response() -> None:
+        async for event in async_generator:
+            # Accessing `ev.property_of_dummy_request` here allows us to validate
+            # mypy has the type information we think it has. We run mypy on the tests.
+            print(event.property_of_dummy_request)
+            nonlocal stream_counter
```
pipermerriam: This should probably be outside of the `for` loop so it only happens once.

tests/core/test_basics.py
pipermerriam: Open an issue in py-evm codebase for this?

tests/core/test_basics.py
```diff
@@ -139,6 +139,40 @@ def expected_response_type() -> Type[DummyResponse]:
...
+    async def cancel_soon() -> None:
+        while True:
+            await asyncio.sleep(0.01)
+            if stream_counter == 2:
+                await async_generator.aclose()
+
+    asyncio.ensure_future(stream_response())
+    asyncio.ensure_future(cancel_soon())
+
+    # we broadcast one more item than what we consume and test for that
```
pipermerriam: the counts seem to not match up with this comment since 50 is not "one more than `2`"

tests/core/test_basics.py
```diff
@@ -156,3 +190,13 @@ def expected_response_type() -> Type[DummyResponse]:
...
     await asyncio.sleep(0.01)
     assert isinstance(received, DummyRequest)
+
+
+@pytest.mark.asyncio
+async def test_wait_for_can_get_cancelled(endpoint: Endpoint) -> None:
+
+    with pytest.raises(asyncio.TimeoutError):
+        await asyncio.wait_for(endpoint.wait_for(DummyRequest), 0.1)
+    await asyncio.sleep(0.1)
```
pipermerriam: Can this get smaller?  `0.1` adds up quickly in tests runs.

tests/core/test_basics.py
cburgdorf: Ah copy and paste bug :)

lahja/endpoint.py
cburgdorf: Ah, sounds good :+1:

tests/core/test_basics.py
pipermerriam: I don't personally find it confusing.  Both of them map to similar concepts...

cburgdorf: Alright, leaving it as is then :+1:

tests/core/test_basics.py
cburgdorf: Yep, checked all sleeps and reduced times.

PR #15, opened 2018-09-27
Implement internal events

### LINKED ISSUE #10: Ensure internal events aren't pushed into central event bus
## What was wrong?

Currently, even if we would broadcast an event only to our very own `Endpoint` it will still go on a roundtrip to the `EventBus` (which usually should sit in another process) which is

a.) a waste of resources

b.) restricts use case to pickable events

If we would short circuit internal events to stay within the `Endpoint` we could get rid of the special legacy `PluginManager` events in Trinity and handle those via the event bus as well. 

## How can it be fixed?

Detect internal events and ensure they stay within the Endpoint.

cburgdorf:
## What was wrong?

As described in #10. Currently the event bus always propagates every event at least one process hop to the process that holds the central `EventBus` to then decide whether to pass it on to a specific `Endpoint` or not.

Not only is this a waste of resources for events that do not need to propagate outside of the current `Endpoint` it will also prevent us from propagating non-pickable events in such cases.

## How was it fixed?

Added `internal` to the `BroadcastConfig` and then decide based on that to either propagate externally or just internally.

#### Cute Animal Picture

### DIFF
--- a/tests/core/test_internal.py
+++ b/tests/core/test_internal.py
@@ -1,0 +1,50 @@
+import asyncio
+from typing import (
+    Any,
+    Tuple,
+)
+
+import pytest
+
+from lahja import (
+    BaseEvent,
+    BroadcastConfig,
+    Endpoint,
+)
+
+
+class DummyResponse(BaseEvent):
+    property_of_dummy_response = None
+
+    def __init__(self, something: Any) -> None:
+        pass
+
+
+@pytest.mark.asyncio
+async def test_internal_propagation(pair_of_endpoints: Tuple[Endpoint, Endpoint]) -> None:
+
+    endpoint1, endpoint2 = pair_of_endpoints
+
+    async def broadcast_dummies() -> None:
+        while True:
+            # We are broadcasting internally on endpoint1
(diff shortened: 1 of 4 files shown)


cburgdorf:
@pipermerriam That's what we talked about briefly in person. This would allow us to refactor the plugin support in Trinity further and get rid of the plugin specific events so that all events are handled via the event bus mechanism.

tests/core/test_internal.py
```diff
@@ -0,0 +1,50 @@
...
+async def test_internal_propagation(pair_of_endpoints: Tuple[Endpoint, Endpoint]) -> None:
+
+    endpoint1, endpoint2 = pair_of_endpoints
+
+    async def broadcast_dummies() -> None:
+        while True:
+            # We are broadcasting internally on endpoint1
+            endpoint1.broadcast(
+                DummyResponse("Dummy"),
+                BroadcastConfig(internal=True)
```
pipermerriam: not for this PR.

It might be nice to be able to shortcut this as `endpoint.broadcast(event, internal=True)`

Issue #17, opened 2018-09-28
Internal events don't work with non-pickable payload


cburgdorf:
## What was wrong?

One of the main motivations for the *internal* events is to be able to broadcast events that stay within one `Endpoint` (and hence `Process`) and allow those to use non-pickable events as well.

Unfortunately, the simple implementation doesn't work in practice. Trying to use this in practice leaves me with:

```
Traceback (most recent call last):
  File "/usr/lib/python3.6/multiprocessing/queues.py", line 234, in _feed
    obj = _ForkingPickler.dumps(obj)
  File "/usr/lib/python3.6/multiprocessing/reduction.py", line 51, in dumps
    cls(buf, protocol).dump(obj)
  File "stringsource", line 2, in uvloop.loop.Loop.__reduce_cython__
TypeError: no default __reduce__ due to non-trivial __cinit__
```

I believe what happens is that even if the event does not intent to move across process boundaries, simply pushing it into a `multiprocessing.Queue` enforces a *pickability* check.

## How can it be fixed?

Probably route these events into a special purpose local queue.

cburgdorf:
This was fixed in #18

PR #18, opened 2018-09-28
Ensure internal events work with non-pickable events

cburgdorf:
## What was wrong?

As described in #17 internal events do not work with non-pickable events as is (but should)

## How was it fixed?

First off, apologies that this turned into some re-architecture that is now mangled in a single commit and hard to detangle. As I was fixing the core problem - which is that internal events need to be propagate through their own special purpose queue - I stumbled over a bunch of problems that needed to be addressed. 

So here is what this does in a nutshell:

- use an `asyncio.Queue` to propagate internal events
- simplify the `def connect()` logic and rename to `def connect_no_wait()`
- add an `async def wait_for_connection()` API that allows async waiting up to the point where the background processing loops have started their work. This fixing an important problem that existed to far which was that one could accidentally `broadcast` before the internal things where actually ready, resulting in run time errors.
- add a new `async def connect()` API that just calls `connect_no_wait` and then does `await self.wait_for_connection`
- ensure `Optional[T]` is handled consistently via non-optional properties that raise a `NoConnection()` error in case `T` happens to be `None`
- turn the `endpoint` / `pair_of_endpoints` fixtures into `async` fixtures ensuring that they do not yield before the endpoints are connected
- migrate a test that previously could not use the `Endpoint` fixture (maybe related to the previous point) to use the fixture.

#### Cute Animal Picture

### DIFF
--- a/lahja/endpoint.py
+++ b/lahja/endpoint.py
@@ -43,68 +43,118 @@ class Endpoint:
         self.name = name
         self._sending_queue = sending_queue
         self._receiving_queue = receiving_queue
-        self._futures: Dict[str, asyncio.Future] = {}
+        self._futures: Dict[Optional[str], asyncio.Future] = {}
         self._handler: Dict[Type[BaseEvent], List[Callable[[BaseEvent], Any]]] = {}
         self._queues: Dict[Type[BaseEvent], List[asyncio.Queue]] = {}
-        self._running = False
         self._executor: Optional[ThreadPoolExecutor] = None
+        self._running = False
+        self._optional_internal_queue: Optional[asyncio.Queue] = None
+        self._optional_loop: Optional[asyncio.AbstractEventLoop] = None
+        self._optional_receiving_loop_running: Optional[asyncio.Event] = None
+        self._optional_internal_loop_running: Optional[asyncio.Event] = None
 
     @property
-    def loop(self) -> asyncio.AbstractEventLoop:
-        if self._loop is None:
-            raise NoConnection("Need to call `connect()` first")
-        return self._loop
+    def _loop(self) -> asyncio.AbstractEventLoop:
+        if self._optional_loop is None:
+            raise NoConnection("Need to connect first")
+        return self._optional_loop
 
-    def connect(self, loop: Optional[asyncio.AbstractEventLoop] = None) -> None:
+    @property
+    def _internal_queue(self) -> asyncio.Queue:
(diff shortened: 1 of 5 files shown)


cburgdorf:
@pipermerriam I validated that this will allow us to make the desired refactoring in Trinity. That said, I think while working on this I realized that our connection logic may have some unsound parts. Basically, I believe the sync `connect()` that schedules a background task may leave us with a race condition which means we can get into a situation where we try to propagate before the endpoint is actually ready.

That's not introduced by this PR and I don't have a immediate solution for this at hand now. I think we can address that with another PR though.

pipermerriam:
Using an `asyncio.Event` is probably the right solution, as that allows `async` code to `await` till the event is set before starting anything.

cburgdorf:
>Using an `asyncio.Event` is probably the right solution

True that! Thanks for pointing it out. Btw, this has grown into a bigger refactoring with a bunch of fixes and improvements. I changed the summary of the PR to elaborate a bit more on the changes and you can see this put to work in https://github.com/ethereum/py-evm/pull/1331.

I like the outcome and if there aren't any issues coming up I'd be happy to cut a new release with these changes.

lahja/endpoint.py
```diff
@@ -43,68 +43,118 @@ def __init__(self,
...
+        self._optional_internal_queue = asyncio.Queue()
 
-    async def _try_connect(self, loop: asyncio.AbstractEventLoop) -> None:
-        # We need to handle exceptions here to not get `Task exception was never retrieved`
-        # errors in case the `connect()` fails (e.g. because the remote process is shutting down)
-        try:
-            await asyncio.ensure_future(self._connect(), loop=loop)
-        except Exception as exc:
-            raise Exception("Exception from Endpoint.connect()") from exc
+        # Using `gather` (over e.g. `wait` or plain `ensure_future`) ensures that the inner futures
```
pipermerriam: That's quite interesting.  Wonder if we can use this elsewhere?

cburgdorf: Yes probably, it's one of the many surprises of Python asyncio ;)

https://docs.python.org/3/library/asyncio-task.html#asyncio.gather

>If `gather()` is cancelled, all submitted awaitables (that have not completed yet) are also cancelled.

Also interesting:

>Unlike `wait_for()`, `wait()` does not cancel the futures when a timeout occurs.

PR #22, opened 2018-11-26
Benchmark

cburgdorf:
Started to put together a benchmark to get some hard numbers about throughput etc. I started to run into weird problems. Just dropping it here to pick it up another day.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -36,6 +36,12 @@ common: &common
         key: cache-{{ .Environment.CIRCLE_JOB }}-{{ checksum "setup.py" }}-{{ checksum "tox.ini" }}
 
 jobs:
+  benchmark:
+    <<: *common
+    docker:
+      - image: circleci/python:3.6
+        environment:
+          TOXENV: benchmark
   doctest:
     <<: *common
     docker:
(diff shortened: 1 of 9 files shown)


cburgdorf:
Found it! It's `fork` vs `spawn` again. The **defafult** method to launch processes on Unix with Python is `fork`. But turns out that `asyncio` just doesn't support that properly and I'm baffled that this isn't more clearly communicated. 

In fact, they may plan [fixing that for Python 3.8](https://bugs.python.org/issue22087#msg318140). 

Here's another issue about [fork support for asyncio](https://bugs.python.org/issue21998#msg325820) with the last comment just two months old.

>Give up, document that fork() is not supported and close the issue :-) IMHO it's not worth it.

So, the problem is that when we use `fork` (default!) we end up with weird shared memory quirks and suddenly have the central `EventBus` (not each `Endpoint`) running in all the connected processes. And from there, madness just takes its turn.

Now that I changed the benchmark to use `spawn` everything works as expected and I don't see any dropped events anymore.

cburgdorf:
Overtaken by https://github.com/ethereum/lahja/pull/23

PR #26, opened 2018-12-18
Do not set results on futures that are cancelled or done

cburgdorf:
## What was wrong?

Currently, it seems it may happen that we run into a situation where we try to resolve a future that was already cancelled. Similarly it could *theoretically* happen that we try to set a result on a future that is already done.
 
This may be a fix for https://github.com/ethereum/py-evm/issues/1613

## How was it fixed?

Check if the future is already done or cancelled before setting the result.

#### Cute Animal Picture

### DIFF
--- a/lahja/endpoint.py
+++ b/lahja/endpoint.py
@@ -145,8 +145,9 @@ class Endpoint:
 
         if in_futures:
             future = self._futures[config.filter_event_id]
-            future.set_result(item)
-            self._futures.pop(config.filter_event_id)
+            if not future.done():
+                future.set_result(item)
+            self._futures.pop(config.filter_event_id, None)
 
         if in_queue:
             for queue in self._queues[event_type]:


cburgdorf:
@pipermerriam This may be a fix for that issue that you've been seeing. I don't have a test for this because I doubt we can test it without raising the complexity of the tests by using real multiple processes (and even then it may be hard to reproduce).

It may also be the case that we actually need some threading locks because `process_item` can end up being processed in multiple threads.

I was thinking maybe you could put this patch to work on your machine since to you seem to be able to reproduce the issue?

lahja/endpoint.py
```diff
@@ -145,8 +145,9 @@ def _process_item(self, item: BaseEvent, config: BroadcastConfig) -> None:
 
         if in_futures:
             future = self._futures[config.filter_event_id]
-            future.set_result(item)
-            self._futures.pop(config.filter_event_id)
+            if not future.cancelled() and not future.done():
```
carver: The check is redundant: `done` includes `cancelled`: https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Future.done

cburgdorf: Ah, right!

lahja/endpoint.py
```diff
@@ -145,8 +145,9 @@ def _process_item(self, item: BaseEvent, config: BroadcastConfig) -> None:
...
+                future.set_result(item)
+                self._futures.pop(config.filter_event_id)
```
pipermerriam: Hrm, wouldn't you still want to do this pop regardless of done/cancelled status?

cburgdorf: Mmh...I don't think so because if we ever get here when the future is already done it will mean that it probably is removed already either because we hit this exact code location before (from another thread) or because we hit this when the future was cancelled.

https://github.com/ethereum/lahja/blob/5e3cd14815df2c279e67bba5dd78475b46741331/lahja/endpoint.py#L227

But now that I think about it, I guess we should play it safe whenever we attempt to remove a key from `self._futures` and check if the key still exists.

In fact, because of the fact that `_process_item` may be called from multiple threads I'm really wondering if we even can avoid using thread locks here. But I would like to try to avoid those if we can.

pipermerriam: You can just do a `self._futures.pop(id, None)` which will work gracefully if the key isn't present.

cburgdorf:
This is blocked on CI failure which has a pending fix https://github.com/ethereum/lahja/pull/27

PR #29, opened 2019-02-04
Add a benchmark that uses Python proxy objects

cburgdorf:
## What was wrong?

We have never compared performance between lahja and a native Python proxy setup yet.

## How was it fixed?

This adds a new script `perf_proxy` that should be roughly comparable to running:

`python scripts/perf_benchmark.py --num-processes 1 --num-events 10000 --payload-bytes 1`

The results are really bad. It shows that the Lahja benchmark performs roughly 15x worse compared to the Proxy solution. There may be some additional overhead involved in the Lahja benchmark but even if we account for that, I guess we'd still be looking at 10x.

This doesn't come as a surprise to me because every event in Lahja travels through one extra process as events are routed through one coordinator process. In addition to that, we can assume that the Python proxy code has already received years of performance tuning whereas Lahja is still pretty much alpha state software.

The main selling point for Lahja at this point isn't efficiency but ease of use as well as enabling an event driven programming model.

There may be many improvements we can do in order to get things more efficient.

#### Cute Animal Picture

### DIFF
--- a/scripts/perf_proxy.py
+++ b/scripts/perf_proxy.py
@@ -1,0 +1,93 @@
+import multiprocessing
+from multiprocessing.managers import (
+    BaseManager,
+    BaseProxy,
+)
+import pathlib
+import time
+
+# This benchmark uses native Python proxy objects to communicate between two processes.
+# With default settings, this should be roughly compareable to:
+# python scripts/perf_benchmark.py --num-processes 1 --num-events 100000 --payload-bytes 1
+
+NUM_CALL_TO_REMOTE = 10000
+PAYLOAD = b'\x00' * 1
+IPC_PATH = 'test.ipc'
+
+class TestThingy:
+
+    def __init__(self):
+        self.i = 0
+
+    def send_bytes(self, val: bytes):
+        # We are doing nothing with the bytes. Just trying to be compatible
+        # to the other benchmark (perf_benchmark.py)
+        self.i += 1
+
+    def get_val(self):
+        return self.i
+
+class ProxyTestThingy(BaseProxy):
+
+    def send_bytes(self, val: bytes) -> None:
+        return self._callmethod('send_bytes', (val,))
+
+    def get_val(self) -> None:
+        return self._callmethod('get_val')
+
+def create_db_server_manager() -> BaseManager:
+
+    class DummyManager(BaseManager):
+        pass
+
+    DummyManager.register(
+        'get_thingy', callable=lambda: TestThingy(), proxytype=ProxyTestThingy)
+
+    manager = DummyManager(address=IPC_PATH)  # type: ignore
+    return manager
+
+
+def create_db_consumer_manager() -> BaseManager:
+    class DummyManager(BaseManager):
+        pass
+
+    DummyManager.register('get_thingy', proxytype=ProxyTestThingy)
+
+    manager = DummyManager(address=IPC_PATH)  # type: ignore
+    manager.connect()
+    return manager
+
+
+def server_proc_launch():
+    manager = create_db_server_manager()
+    print("server process launched")
+    server = manager.get_server()
+    server.serve_forever()
+
+def proc_launch():
+    manager = create_db_consumer_manager()
+    print("consumer process launched")
+    thingy = manager.get_thingy()
+
+    back_then = time.perf_counter()
+    for i in range(NUM_CALL_TO_REMOTE):
+        thingy.send_bytes(PAYLOAD)
+    now = time.perf_counter()
+    duration = now - back_then
+    print(f"we are at {thingy.get_val()}, duration: {duration}")
+
+if __name__ == "__main__":
+
+    print("starting proxy server")
+    
+    server_proc = multiprocessing.Process(
+        target=server_proc_launch,
+    )
+    server_proc.start()
+
+    print("starting consumer process")
+
+    consumer_proc = multiprocessing.Process(
+        target=proc_launch,
+    )
+    consumer_proc.start()


pipermerriam:
@cburgdorf can we account for the extra process and benchmark proxy object communications that cross two process boundaries?

My thought here is that the *current* lahja API is really optimized for this broadcast style model.  I'm wondering if we should look into a second abstraction for direct `1:1` communication.

cburgdorf:
>can we account for the extra process and benchmark proxy object communications that cross two process boundaries?

I could change the code so that it also goes through two processes, bringing it closer to the current lahja model.

>My thought here is that the current lahja API is really optimized for this broadcast style model. I'm wondering if we should look into a second abstraction for direct 1:1 communication.

@pipermerriam I'm currently looking into options. I believe it should be possible to get rid of that extra process jump.

cburgdorf:
@pipermerriam I hacked together a PoC that gets rid of the central coordinator. It's an early start but I actually think this can turn out quite well.

https://github.com/ethereum/lahja/pull/30/

pipermerriam:
@cburgdorf have you done any actual profiling of the two approaches to see where the time is being spent?

cburgdorf:
@pipermerriam I haven't yet but I will do that. But without digging too deep in, the obvious difference is:

### A. Current Lahja:

1. An event is broadcasted within an `Endpoint` in process X

2. It is pushed into a `multiprocessing.Queue` which is shared between the process `X` and the `main` process. The actual transport of the data is handled internally by the `multiprocessing.Queue` which I assume uses roughly the same primitives that a proxy object uses to call out to another process.

3. Then, in the `main` process, we use a thread pool to read from the otherwise blocking `multiprocessing.Queue.get` API. At this point, we are inside the event loop of the `main`  process.

https://github.com/ethereum/lahja/blob/049bf778b68189a5fc7c1b55f1e9325b769a3f2c/lahja/eventbus.py#L62-L73

4. The data is pushed into the `multiprocessing.Queue` of the `Endpoint` in process `Y`. Again, the `multiprocessing.Queue` takes care of transferring the actual data.

5. Then, in the `Y` process, we use a thread pool again to read from the otherwise blocking `multiprocessing.Queue.get` API. At this point, we are inside the event loop of the `Y`  process.

6. We then feed it into an `asyncio.Queue`, a `Future` or straight into a `Callable` depending on how the event is subscribed to.

### B. In the POC code

1. An event is broadcasted within an `Endpoint` in process X

2. It is pushed into a `Proxy` that is directly connected to an `asyncio.Queue` in process `Y`. So, at this point, we are already in the `EventLoop` of process `Y`.

3. From there, the data is fed it into an `asyncio.Queue`, a `Future` or straight into a `Callable` depending on how the event is subscribed to.

So, not only do we eliminate the `main` process that Lahja currently uses as a middleman to connect all processes, we also do not rely on any `multiprocessing.Queue` anymore, which also means, we don't need to read from those via thread pool. So, this whole `async_get` util isn't needed with this approach.

Again, I believe that the underlying mechanics of how a `multiprocessing.Queue` can be shared across different processes aren't that much different from those that are used within the `BaseProxy`. The biggest efficiency gain can be expected to come from getting rid of that extra process. However, I could imagine that there is some extra overhead involved in using a `multiprocessing.Queue` and reading from it through a thread pool, compared to simply pushing data through a `BaseProxy` straight into an `asyncio.Queue`.

I would actually be surprised if you could get much faster than `BaseProxy` access with raw stdlib Python. You can obviously with 3rd party libs such as `gRPC` and I believe Lahja could also use these as its underlying transportation mechanism.

pipermerriam:
> I believe that the underlying mechanics of how a `multiprocessing.Queue` can be shared across different processes aren't that much different from those that are used within the `BaseProxy`.

This is accurate based on my memory of reading a bunch of the `multiprocessing` internals and implementation details a while back.

> I would actually be surprised if you could get much faster than `BaseProxy`

If we assume we have the same amount of raw cross-process communication overhead then my memory suggests that we can still cut out a decent bit of extra logic handling from the `BaseProxy` implementation, but if the cross-process communication accounts for 95% of the time then that extra logic overhead doesn't really matter.  Basically, I do recall that the `multiprocessing.Manager` does a decent amount of juggling using various queues and locks internally.

cburgdorf:
>does a decent amount of juggling using various queues and locks internally.

Of which we don't know how much is needed to ensure correctness (e.g. concurrent calls into the proxy). 
It's worth taking a closer look but I briefly looked into `BaseProxy` and my intuition says I'm skeptical that we can outsmart what is there :)

It would also be interesting to see how much the performance could be improved if the underlying transport would be handled through `gRPC`.

Btw, I just benchmarked the new Lahja PoC against the native `BaseProxy` setup and now both solutions are roughly in the same ballpark. The raw `BaseProxy` solution is still 1.45x faster which is logical given that the Lahja PoC solution is just extra ergonomics on top of the same underlying transport layer. However, in practice that extra ergonomics can buy you some perf advantage back depending on the concrete circumstances I believe (IO boundness etc).

Btw, getting rid of that central coordinator process comes with another huge advantage: It makes joining of new Endpoints at runtime really easy. With the previous setup, it was kind of mandatory to bootstrap all Endpoints at startup (unless some funky tricks are developed) whereas in this PoC, connecting to other Endpoints can be done at any time.

I'm trying to polish things up and see if I can move this from a PoC to a proper implementation that passes tests and supports all the existing benchmarks.

cburgdorf:
Looking more deeply into this, I know believe that connecting the processes via [`asyncio streams`](https://docs.python.org/3/library/asyncio-stream.html) would be the best case scenario.

The reason for that is, when we call a method on a `BaseProxy` we are causing IO in a synchronous blocking manner. Using something like `loop.run_in_executor` and running it through a`ThreadPoolExecutor` doesn't help here (in fact, measuring shows it makes it worse) since after all `asyncio` is mainly a tool to do IO *without* threads leveraging OS APIs that can notify us when resources become available.

This leads me to believe, that the best case scenario may in fact involve using the low level asyncio stream primitives.

But I'll push this further down the road. Just changing the topology and using `BaseProxy` for the communication is already such a major speed improvement and at least moves us into a region where it should be safe to use Lahja for more frequent events such as those in the Trinity PeerPool.

pipermerriam:
:+1: 

At some point it'd be cool to do a **really** rough prototype of this using the asyncio stream primitives.  My intuition says that it should be pretty straight forward to implement a rough equivalent of `multiprocessing.Pipe` that allows passing objects across boundaries.

PR #30, opened 2019-02-04
Get rid of extra process hop

cburgdorf:
## What was wrong?

Currently, Lahja does one extra process hop for every transmitted event

## How was it fixed?

This gets rid of the extra hop by creating a different topology in which every endpoint is directly connected to every other endpoint (fine grained configuration possible)

Some things to highlight:

- it's a huge performance difference (see benchmark. Especially avg propagation time which was known to be poor when massively flooding with events)
- we get rid of the `EventBus` class entirely. This class was previously used for the manager process that connected all the endpoints together. We could think about if we want to rename `Endpoint` to `EventBus` or maybe to `EventbusEndpoint` or something like that?
- we get rid of some additional threading overhead
- `Endpoint`s are no longer created in a "manager process" and then passed into another process. They are directly created in the process that wants to control it.
- `filter_endpoint` got a lot more resource friendly because it now means an event that is broadcasted using `filter_endpoint='foo'` does really only travel to that single endpoint `foo` whereas before it at least traveled to the manager process.
- this also fixes the issue with the benchmark where it previously would not cleanly shut down.

**I'm feeling mildly convinced that we can get more efficient by using some lower level `asyncio` primitives instead of `BaseProxy` but at the same time I feel that this gives us enough efficiency to at least carry on with my original task which was blocked by a too poorly performing EventBus.**

## Benchmarks

Benchmarks look really promising. Here's a benchmark where the current Lahja implementation can barely keep up and avg propagation time can crawl up as high as 7 seconds.

```
$ python ./scripts/perf_benchmark.py --num-processes 10 --num-events 10000
                                                                    +++Globals+++                                                                     
|        10         |    15.93889    |     789521.29491      |       10000        |     100000     |    627.396     |    6273.964    |
                                                                +++Process Details+++                                                                 
|    consumer_7     |     10000      |  11:20:50.632  |  11:21:06.471  |    0.01416     |    15.58441    |    7.89532     |    15.83871    |      78953.22804      |
|    consumer_4     |     10000      |  11:20:50.632  |  11:21:06.472  |    0.01286     |    15.58499    |    7.89512     |    15.83930    |      78951.22536      |
|    consumer_2     |     10000      |  11:20:50.632  |  11:21:06.471  |    0.01265     |    15.58421    |    7.89516     |    15.83851    |      78951.56178      |
|    consumer_1     |     10000      |  11:20:50.632  |  11:21:06.477  |    0.01175     |    15.59063    |    7.89512     |    15.84494    |      78951.22707      |
|    consumer_8     |     10000      |  11:20:50.632  |  11:21:06.477  |    0.01831     |    15.59056    |    7.89535     |    15.84486    |      78953.48364      |
|    consumer_3     |     10000      |  11:20:50.632  |  11:21:06.476  |    0.01279     |    15.58923    |    7.89511     |    15.84353    |      78951.07581      |
|    consumer_9     |     10000      |  11:20:50.632  |  11:21:06.471  |    0.01831     |    15.58430    |    7.89543     |    15.83860    |      78954.31422      |
|    consumer_5     |     10000      |  11:20:50.632  |  11:21:06.472  |    0.01350     |    15.58499    |    7.89519     |    15.83930    |      78951.91001      |
|    consumer_6     |     10000      |  11:20:50.632  |  11:21:06.482  |    0.01402     |    15.59568    |    7.89521     |    15.84998    |      78952.06387      |
|    consumer_0     |     10000      |  11:20:50.632  |  11:21:06.470  |    0.01175     |    15.58373    |    7.89512     |    15.83803    |      78951.20510      |
```

The same with the new upcoming Lahja version that gets rid of the central coordinator process topology:

```
$ python ./scripts/perf_benchmark.py --num-processes 10 --num-events 10000
                                                                    +++Globals+++                                                                     
|        10         |    12.06595    |       72.18670        |       10000        |     100000     |    828.779     |    8287.788    |
                                                                +++Process Details+++                                                                 
|    consumer_2     |     10000      |  11:20:19.427  |  11:20:31.331  |    0.00032     |    0.00834     |    0.00044     |    11.90361    |        4.35764        |
|    consumer_0     |     10000      |  11:20:19.427  |  11:20:31.333  |    0.00014     |    0.00318     |    0.00020     |    11.90554    |        1.95367        |
|    consumer_1     |     10000      |  11:20:19.427  |  11:20:31.330  |    0.00023     |    0.00414     |    0.00031     |    11.90293    |        3.11184        |
|    consumer_8     |     10000      |  11:20:19.427  |  11:20:31.359  |    0.00088     |    0.02853     |    0.00113     |    11.93111    |       11.32660        |
|    consumer_3     |     10000      |  11:20:19.427  |  11:20:31.331  |    0.00041     |    0.00553     |    0.00054     |    11.90322    |        5.42828        |
|    consumer_5     |     10000      |  11:20:19.427  |  11:20:31.343  |    0.00059     |    0.01333     |    0.00078     |    11.91591    |        7.78999        |
|    consumer_4     |     10000      |  11:20:19.427  |  11:20:31.343  |    0.00050     |    0.01321     |    0.00067     |    11.91579    |        6.66338        |
|    consumer_7     |     10000      |  11:20:19.427  |  11:20:31.364  |    0.00079     |    0.03361     |    0.00101     |    11.93619    |       10.12145        |
|    consumer_9     |     10000      |  11:20:19.427  |  11:20:31.370  |    0.00098     |    0.04037     |    0.00125     |    11.94295    |       12.46429        |
|    consumer_6     |     10000      |  11:20:19.427  |  11:20:31.358  |    0.00070     |    0.02821     |    0.00090     |    11.93080    |        8.96957        |
```

Every event is still propagated fast no matter how hard we flood the system.

#### Cute Animal Picture


cburgdorf:
There's also a WIP PR for Trinity that puts that version to work https://github.com/ethereum/trinity/pull/263

cburgdorf:
@pipermerriam I've already made the change in Trinity to adapt to this PR. It seems to be working great so far.

https://github.com/ethereum/trinity/pull/263

cburgdorf:
@pipermerriam I think I will soon have a PR in Trinity ready that will move the request server into its own plugin but it builds on a bunch of PRs and this one is the root of it. So, whenever you have some time to review stuff it would be good starting with this PR. While we certainly have lots of room to improve the EventBus, from what I see so far, with this PR it seems scalable enough to use it for what we want.

lithp:
> it's a huge performance difference (see benchmark. Especially avg propagation time which was known to be poor when massively flooding with events)

This is an impressive improvement! Do we know what was causing it to be so slow before? If we're now only doing one hop instead of two hops, why is the performance so much better than double what it used to be? The "fastest" column is now something 44x times faster, I wouldn't have expected such a large improvement.

lahja/_utils.py
```diff
@@ -0,0 +1,26 @@
+import asyncio
+import pathlib
+import time
+
+
+def wait_for_path_blocking(path: pathlib.Path, timeout: int=30) -> None:
```
lithp: It's sad there isn't a cross-platform way to do this :cry: 

On linux you have [inotify](http://man7.org/linux/man-pages/man7/inotify.7.html) but that requires [a dependency](https://pypi.org/project/inotify/), it looks like [libfswatch](https://github.com/emcrisostomo/fswatch) abstracts the platform away from you but I can't find any python wrapper.

lahja/_utils.py
```diff
@@ -0,0 +1,26 @@
...
+    ``path`` otherwise raises :exc:`TimeoutError`.
+    """
+    start_at = time.monotonic()
+    while time.monotonic() - start_at < timeout:
+        if path.exists():
+            return
+        else:
+            time.sleep(0.05)
+
+    raise TimeoutError("IPC socket file has not appeared in %d seconds!" % timeout)
```
lithp: For easier debugging maybe include the path which was being polled?

lahja/eventbus.py
```diff
@@ -1,82 +0,0 @@
-import asyncio
```
lithp: It feels good to delete so much code :)

lahja/endpoint.py
```diff
@@ -33,62 +38,102 @@
 )
 
 
+class ConnectionConfig(NamedTuple):
```
lithp: After looking through the tests, it feels like requiring consumers to pass `ConnectionConfig`s rather than names imposes a large cost on them for what is essentially an implementation detail (which Path to connect to). Maybe you expect to add more details to `ConnectionConfig`s in the future? If `Endpoints` will usually be referred to by names, what do you think of passing those names directly into methods like `connect_no_wait` and having them derive the appropriate Path themselves?

cburgdorf:
>Do we know what was causing it to be so slow before?

It's multiple things, I broke it down in [this comment](https://github.com/ethereum/lahja/pull/29#issuecomment-460445971) here.

There's still room for another big improvement if we can [tackle this](https://stackoverflow.com/questions/54636691/how-to-efficiently-use-asyncio-when-calling-a-method-on-a-baseproxy). Not only here in Lahja but also in Trinity for every single DB access. Btw, if you don't mind, upvote the question for visibility. Or feel free to take a stab on a solution. I think this could be a big improvement.

lahja/endpoint.py
cburgdorf: >what is essentially an implementation detail (which Path to connect to)

It is for many cases which is why `from_name` exists to make it smoother for the happy path where you don't care about the ipc path. However, e.g. in Trinity we want all IPC paths to go into a certain directory.

lahja/_utils.py
cburgdorf: >It's sad there isn't a cross-platform way to do this

Wait, but [`pathlib.Path`](https://docs.python.org/3/library/pathlib.html) should work cross-platform or are we talking about different things?

lahja/endpoint.py
```diff
@@ -33,62 +38,102 @@
...
-        if self._optional_internal_loop_running is None:
-            raise NoConnection("Need to connect first")
-        return self._optional_internal_loop_running
-
-    def connect_no_wait(self, loop: Optional[asyncio.AbstractEventLoop] = None) -> None:
+    def connect_no_wait(self,
+                        connection_config: ConnectionConfig,
+                        loop: Optional[asyncio.AbstractEventLoop] = None) -> None:
         """
         Connect the :class:`~lahja.endpoint.Endpoint` to the :class:`~lahja.eventbus.EventBus`
```
lithp: This comment is out of date

lahja/_utils.py
lithp: I mean the spinloop to check whether the file exists, `inotify` would let you create a file descriptor which you could `read` in order to block until an event (the file was created) occurred. It would block the thread without eating up CPU time.

lahja/endpoint.py
lithp: Gotcha, that makes sense.

Is there another way to make this interface simpler? Maybe something like:

```python
# the caller keeps track of this
ctx = lahja.ConnectionContext('/tmp/ipc_dir') 

# so that later it can call this:
endpoint = ctx.Endpoint('peerpool')
```

Callers would lose control over the exact layout of the ipcs but would still be able to specify a directory to use and Endpoints would be very easy to create and reference. This would also allow for something like a `MemoryContext` for running quick tests without creating ipc files.

I don't know, that sounds like it has its own drawbacks. Definitely feel free to ignore me but it feels heavyweight to specify the name and exact path every time you want to reference an Endpoint when one could be used to derive the other.

cburgdorf: We could probably keep track of a directory as you propose but it would still give up a bit flexibility. Different endpoints may reside in entirely different directories. I'm not sure if we really want to give up on that flexibility.

>it feels heavyweight to specify the name and exact path every time you want to reference an Endpoint

Which really you don't need if you don't care about the path. Then you just use `ConnectionConfig.from_name(`foo`)` and it will derive it on it's own.

But I guess we could come up with a scheme that allows both fine-grained case by case as well as directory based path resolution.

But considering that most applications will only need these connect calls in very few places it doesn't catch me that strongly.

lahja/_utils.py
cburgdorf: Ah! Gotcha. Yes, I was quite surprised not to find something like that in the stdlib.

lahja/endpoint.py
```diff
@@ -129,6 +175,65 @@ def connect_no_wait(self, loop: Optional[asyncio.AbstractEventLoop] = None) -> N
 
             self._process_item(item, config)
 
+    def _create_external_api(self, ipc_path: pathlib.Path) -> None:
+
+        receiver = EndpointConnector(self)
+
+        ConnectorManager = get_connector_manager()
```
lithp: What do you think of making the flow a little simpler by in-lining this method? By in-lining it some code is duplicated but I think it makes the code easier to read. At first glance `get_connector_manager` looks like it returns a global instance, but looking at the definition reveals it's creating a fresh class each time.

cburgdorf: Yep, I was questioning the standalone function as well. Either inlining or just a private method on the endpoint itself. Which would you prefer?

lithp: I think that I'd prefer in-lining, the docs give me the impression that creating the class and calling `.register` is a single operation, it'd be nice to have them be next to each other.

cburgdorf: Sounds good!

lahja/endpoint.py
```diff
@@ -129,6 +174,65 @@ def connect_no_wait(self, loop: Optional[asyncio.AbstractEventLoop] = None) -> N
...
+
+        ConnectorManager.register('get_connector', callable=lambda: receiver)  # type: ignore
+
+        manager = ConnectorManager(address=str(ipc_path))  # type: ignore
+        server = manager.get_server()   # type: ignore
+        threading.Thread(target=server.serve_forever, daemon=True).start()
+
+    def connect_to_endpoints_blocking(self, *endpoints: ConnectionConfig, timeout: int=30) -> None:
```
lithp: I think you might not need this method, and transitively you might not need `wait_for_path_blocking`.

Every caller of `connect_to_endpoints_blocking` has an EventLoop, and could instead call:

```
loop.run_until_complete(endpoint.connect_to_endpoints(...))
```

This would even give a small performance boost, because `connect_to_endpoints` works in parallel.

lahja/endpoint.py
```diff
@@ -33,62 +38,101 @@
...
+        self._ipc_path = connection_config.path
+        self._create_external_api(self._ipc_path)
+        self._loop = loop
+        self._internal_loop_running = asyncio.Event(loop=self._loop)
+        self._receiving_loop_running = asyncio.Event(loop=self._loop)
+        self._internal_queue = asyncio.Queue(loop=self._loop)
+        self._receiving_queue = asyncio.Queue(loop=self._loop)
 
         # Using `gather` (over e.g. `wait` or plain `ensure_future`) ensures that the inner futures
         # are automatically cancelled as soon as the parent task is cancelled
```
lithp: Not part of this PR but the lines below this confuse me:

```python
asyncio.gather(
    asyncio.ensure_future(self._connect_receiving_queue(), loop=self._loop),
    asyncio.ensure_future(self._connect_internal_queue(), loop=self._loop),
    loop=self._loop
)
```

Doesn't this cause a warning? I think that `ensure_future` schedules these tasks to be run, but asyncio.gather() creates a task which is never awaited.

lahja/endpoint.py
```diff
@@ -129,6 +174,65 @@ def connect_no_wait(self, loop: Optional[asyncio.AbstractEventLoop] = None) -> N
...
+        established. Raises a ``TimeoutError`` if connections do not become available within
+        ``timeout`` seconds (default 30 seconds).
+        """
+        for endpoint in endpoints:
+            wait_for_path_blocking(endpoint.path, timeout)
+            self._connect_to_endpoint(endpoint)
+
+    async def connect_to_endpoints(self, *endpoints: ConnectionConfig) -> None:
+        """
+        Asynchronously connect to the given endpoints and await until all connections are
```
lithp: I don't think this counts as asynchronous, it blocks the caller until it's finished connecting!

lahja/endpoint.py
```diff
@@ -129,6 +174,65 @@ def connect_no_wait(self, loop: Optional[asyncio.AbstractEventLoop] = None) -> N
...
+        ConnectorManager.register(  # type: ignore
+            'get_connector',
+            proxytype=ProxyEndpointConnector
+        )
+
+        manager = ConnectorManager(address=str(endpoint.path))  # type: ignore
+        manager.connect()
+        self._connected_endpoints[endpoint.name] = manager.get_connector()  # type: ignore
+
+    def is_connected_to(self, endpoint_name: str) -> bool:
```
lithp: This (and `_connected_endpoints` in general) feels like a minor footgun. Because the caller has control of (name, path) there are no guarantees that names are unique. Maybe `_connected_endpoints` should use `(name, path)` as the key? Or maybe the `connect_to_endpoints_*` methods should check that the name doesn't already exist and throw an error if it does.

lahja/endpoint.py
```diff
@@ -100,8 +144,11 @@ def connect_no_wait(self, loop: Optional[asyncio.AbstractEventLoop] = None) -> N
 
         self._running = True
 
-    async def connect(self, loop: Optional[asyncio.AbstractEventLoop] = None) -> None:
-        self.connect_no_wait(loop)
+    async def connect(self,
```
lithp: The `connect_*` methods and the `connect_to_endpoints_*` methods aren't very visually distinct, it might be easy to accidentally mistake one for the other. Maybe they could be renamed to something further apart?

- `start_serving` and `connect_to_endpoints`
- `serve_local_endpoint` and `connect_remote_endpoints`

lahja/endpoint.py
```diff
@@ -168,9 +272,6 @@ def stop(self) -> None:
         self._running = False
         self._receiving_queue.put_nowait((TRANSPARENT_EVENT, None))
         self._internal_queue.put_nowait((TRANSPARENT_EVENT, None))
-        if self._executor is not None:
-            self._executor.shutdown()
-        self._receiving_queue.close()
```
lithp: I don't think this `stop` method prevents messages from being sent to this EventBus. Once `_running` is False the `_connect_receiving_queue` will stop running, but messages sent to this Queue will continue to be sent and will queue up forever.

lahja/endpoint.py
cburgdorf: Ha! That's a really good point! Eagle eyes!

lahja/endpoint.py
cburgdorf: I remember I had put it there to get rid of a cancel that I was seeing in Trinity when I did not have it. The point about `gather` is that it ensures all child tasks are cancelled when the gather task is cancelled.

From the docs: https://docs.python.org/3/library/asyncio-task.html#asyncio.gather
>If gather() is cancelled, all submitted awaitables (that have not completed yet) are also cancelled.

But I see what you mean, it looks weird. Definitely worth taking another look!

lahja/endpoint.py
cburgdorf: Well there's blocking and blocking. This blocks the caller in the same way `await asyncio.sleep(5)` "blocks" the caller but it does not block the thread and consequently not the event loop.

How about just:

>Connect to the given endpoints and await until all connections are established.

lahja/endpoint.py
cburgdorf: Ah wait. I tried this and run into problems. Because, we could only do `loop.run_until_complete` if this loop did not start already. Which is an assumption that doesn't hold because you should be able to connect to further endpoints at any time.

lahja/endpoint.py
cburgdorf: >Or maybe the `connect_to_endpoints_*` methods should check that the name doesn't already exist and throw an error if it does.

I added that. Because, The currently Lahja always works under the assumption that names are unique (and ensuring that got lost as I dropped `EventBus`). E.g. you can use `BroadcastConfig(filter_endpoint="foo")` to only propagate something to the `foo` endpoint.

lahja/endpoint.py
cburgdorf: >* `start_serving` and `connect_to_endpoints`

Let's do this!

lahja/endpoint.py
cburgdorf: Ah yes, the logic there changed quite a bit because there's no `close()` for `asyncio.Queue`. But I addressed this now at the level where we try to push into the queue. I'm raising a warning if we attempt to push into an Endpoint that isn't running yet/anymore.

cburgdorf:
Hey @lithp really great feedback! I tried to address your points. If you feel like tweaking things you should also fee free to push into this branch. And if you feel we are good to go make sure to *approve* the PR :)

lahja/endpoint.py
lithp: Oops :( Thanks for trying it! I should have tried it before making the suggestion. There might be a solution like:

```
if self._loop.running():
   ...
...
```

but that doesn't sound any better than leaving this method as-is :+1:

lahja/endpoint.py
lithp: Oh, interesting. I don't think the gather will ever be cancelled, since it's never scheduled, but maybe this binds together the fates of these two coros? If `_connect_receiving_queue` is canceled then `_connect_internal_queue` will be as well?

Thanks for the pointer, I'll play with it sometime

lahja/endpoint.py
lithp: Sounds good!

lithp:
Hey, thanks, and sorry for taking so long to get to and to review this! I know you're blocked on it, about to get onto a flight but I'll finish up as soon as I can. Excited to see this merged, looks like an amazing improvement!

lahja/endpoint.py
```diff
@@ -33,62 +41,104 @@
...
 
+class ConnectionConfig(NamedTuple):
+    """
+    Configuration class needed to establish :class:`~lahja.endpoint.Endpoint` connections.
+    """
+    name: str
+    path: pathlib.Path
+
+    @staticmethod
+    def from_name(name: str) -> 'ConnectionConfig':
```
pipermerriam: You could give *optional* control over the base path by allowing a `Path` as an optional second argument.

```python
def from_name(name, base_path=None):
    if base_path is None:
        return ConnectionConfig(...)
    elif base_path.is_dir():
        return ConnectionConfig(name=name, path = base_path / name)
    else:
        raise TypeError("Something something bad path explanation")
```

lahja/endpoint.py
```diff
@@ -33,62 +41,104 @@
...
 
 
+class ConnectionConfig(NamedTuple):
+    """
+    Configuration class needed to establish :class:`~lahja.endpoint.Endpoint` connections.
+    """
+    name: str
+    path: pathlib.Path
+
+    @staticmethod
```
pipermerriam: Why `staticmethod`.  Seems like `classmethod` would be more appropriate since you could use `cls(...)` for the instantiation.

lahja/tools/benchmark/utils/config.py
```diff
@@ -0,0 +1,17 @@
...
+)
+
+from lahja import (
+    ConnectionConfig,
+)
+
+
+def create_consumer_endpoint_configs(num_processes: int) -> Tuple[ConnectionConfig, ...]:
+    return tuple(
+        ConnectionConfig.from_name(create_consumer_endpoint_name(i)) for i in range(num_processes)
```
pipermerriam: Minor but it seems simpler to just inline the f-string:

```python
ConnectionConfig.from_name(f"consumer_{i}")
```

lahja/endpoint.py
cburgdorf: Looks useful :+1:

lahja/endpoint.py
cburgdorf: Ah yep. Just to make sure I understand the reasoning (part of me is still a Python noob), using `cls`would ensure it works for subclasses as well, right?

lahja/tools/benchmark/utils/config.py
cburgdorf: Yes but we use that at another place and I wouldn't want two places to maintain a `f"consumer_{i}" thing.

lahja/endpoint.py
```diff
@@ -100,26 +150,45 @@ def connect_no_wait(self, loop: Optional[asyncio.AbstractEventLoop] = None) -> N
...
+        self.start_serving_nowait(connection_config, loop)
+        await self.wait_until_serving()
 
-    async def wait_for_connection(self) -> None:
+    async def wait_until_serving(self) -> None:
         """
-        Wait until the ``Endpoint`` has established a connection to the ``EventBus``
+        Await until the ``Endpoint`` is ready to receive events.
         """
         await asyncio.gather(
```
lithp: nitpick: I think that the UX has gotten slightly worse here. Previously this method would raise `NoConnection` if this method was called too soon. Now it'll throw an `AttributeError`, because `_receiving_loop_running` and `_internal_loop_running` don't exist yet.

lahja/endpoint.py
```diff
@@ -129,6 +198,68 @@ def connect_no_wait(self, loop: Optional[asyncio.AbstractEventLoop] = None) -> N
...
+        self._throw_if_already_connected(*endpoints)
+        for endpoint in endpoints:
+            wait_for_path_blocking(endpoint.path, timeout)
+            self._connect_to_endpoint(endpoint)
+
+    async def connect_to_endpoints(self, *endpoints: ConnectionConfig) -> None:
+        """
+        Connect to the given endpoints and await until all connections are established.
+        """
+        self._throw_if_already_connected(*endpoints)
```
lithp: nitpick: This works in `connect_to_endpoints_blocking` but there's a race condition in `connect_to_endpoints`. I think that something like this would succeed even though it shouldn't:

```python
asyncio.gather(
    endpoint.connect_to_endpoints([ConnectionConfig.from_name('hi', path="first_path.ipc")]),
    endpoint.connect_to_endpoints([ConnectionConfig.from_name('hi', path="second_path.ipc")])
)
```

Because `connect_to_endpoints` does the check but defers actually adding the endpoint there's a race condition where a second endpoint with the same name could get past the check before the first one is added.

lahja/endpoint.py
```diff
@@ -129,6 +198,68 @@ def connect_no_wait(self, loop: Optional[asyncio.AbstractEventLoop] = None) -> N
...
+            loop=self._loop
+        )
+
+    def connect_to_endpoints_nowait(self, *endpoints: ConnectionConfig) -> None:
+        """
+        Connect to the given endpoints as soon as they become available but do not block.
+        """
+        self._throw_if_already_connected(*endpoints)
+        for endpoint in endpoints:
+            asyncio.ensure_future(self._await_connect_to_endpoint(endpoint))
```
lithp: Same here as in `connect_to_endpoints`, there's a race-condition between checking whether the name already exists and then claiming the name.

lahja/endpoint.py
```diff
@@ -185,7 +312,11 @@ def broadcast(self, item: BaseEvent, config: Optional[BroadcastConfig] = None) -
             # and are directly put into the local receiving queue instead.
             self._internal_queue.put_nowait((item, config))
         else:
-            self._sending_queue.put_nowait((item, config))
+            # Broadcast to every connected Endpoint that is allowed to receive the event
+            for name, connector in self._connected_endpoints.items():
+                verboten = config is not None and not config.allowed_to_receive(name)
```
lithp: nitpick: as much as I love German (:smile:) the double-negative is a little hard for me to read. Something like:

```python
allowed = (config is None) or config.allowed_to_receive(name)
if allowed:
   ...
```

might be easier to read.

lahja/endpoint.py
cburgdorf: Yep, I noticed that, too. I will address this before merging.

lahja/endpoint.py
cburgdorf: Yep, I also noticed this. In fact, if you look at the test, I left out the one for the `nowait` API because this one is currently affected by this. I *think* the problem here is that in the affected cases you might end up with the exception thrown in a background task which is kinda bad. But maybe duplicating the check there is still the right thing to do to at least not sweep it under the rug.

lahja/endpoint.py
cburgdorf: Yes, that looks better. :+1:  Are the parenthesis only for better logical grouping or are they actually needed?

cburgdorf:
>Let me know if I'm not giving you the kind of reviews that you're expecting, I don't know whether I'm diving too deep into these changes.

It is super helpful and valuable. I've been working pretty much alone on this lib so far and I'm very grateful to have more eyes on it. Also, I will admit that I can be kind of sloppy with things when I just want to get *this one thing* done :sweat_smile: 

In this case, I *really* want to rip the `networking` process in Trinity apart (into three processes as isolated plugins) and so this is my main concern which kinda makes me blind (or careless) for some of the things you pointed out.

This isn't necessarily bad and often I think it is ok to just create issues for some the things that are ok to address at a later stage. I try to be pragmatic, leave code in a better state than I found it, we'll never be finish with it anyway :)

So, again, very happy with your deep dive review :pray: :+1:

PR #32, opened 2019-02-18
Don't throw but warn in background task

cburgdorf:
## What was wrong?

Throwing exceptions in a background task is annoying to handle. This is a rare case that only happens when we are still waiting on a connection while we try to connect to the same endpoint again.

## How was it fixed?

Just issue a warning rather than throwing.

#### Cute Animal Picture

### DIFF
--- a/lahja/endpoint.py
+++ b/lahja/endpoint.py
@@ -233,7 +233,7 @@ class Endpoint:
         self._throw_if_already_connected(*endpoints)
         for endpoint in endpoints:
             wait_for_path_blocking(endpoint.path, timeout)
-            self._connect_to_endpoint(endpoint)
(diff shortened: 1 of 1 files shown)


lahja/endpoint.py
```diff
@@ -258,10 +258,13 @@ def connect_to_endpoints_nowait(self, *endpoints: ConnectionConfig) -> None:
         self._connect_to_endpoint(endpoint)
 
     def _connect_to_endpoint(self, endpoint: ConnectionConfig) -> None:
```
pipermerriam: Maybe naming this `_connect_if_not_already_connected` would make this functionality more explicitly clear when reading the code in places where it's used.

cburgdorf: Sounds good to me!

PR #33, opened 2019-02-25
Fix issue with logger not set up properly

cburgdorf:
## What was wrong?

The logger wasn't setup correctly, it did not use `logging.getLogger(...)`

## How was it fixed?

Properly setup logger.

#### Cute Animal Picture

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -15,7 +15,7 @@ extras_require = {
     ],
     'lint': [
         "flake8==3.4.1",
-        "isort>=4.2.15,<5",
+        "isort==4.3.4",
         "mypy==0.620",
     ],
     'doc': [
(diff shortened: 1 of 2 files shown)


setup.py
```diff
@@ -15,7 +15,7 @@
     ],
     'lint': [
         "flake8==3.4.1",
-        "isort>=4.2.15,<5",
+        "isort==4.3.4",
```
cburgdorf: A newer `isort` version rejects the trailing comma for multi line imports and the advertised `--trailing-comma` doesn't seem to work either. Pinning to the last version that works for now.

cburgdorf:
@lithp This fixes the lahja logging issues in Trinity.

PR #35, opened 2019-03-18
Use snappy compression if available

cburgdorf:
## What was wrong?

Lahja doesn't currently use any form of compression. This can become a performance bottleneck when sending large events (e.g. propagating blocks, receipts, transactions etc in Trinity).

See benchmarks:

1. Current `master`

```
$ python scripts/perf_benchmark.py --num-processes 1 --num-events 100 --payload-bytes 50000000
                                                                    +++Globals+++                                                                     
|         1         |    9.71082     |        9.80165        |        100         |      100       |     10.298     |     10.298     |
                                                                +++Process Details+++                                                                 
|    consumer_0     |      100       |  11:47:13.169  |  11:47:22.797  |    0.07634     |    0.13488     |    0.09802     |    9.62837     |        9.80165        |
```

2. This PR without `python-snappy` installed

```
$ python scripts/perf_benchmark.py --num-processes 1 --num-events 100 --payload-bytes 50000000
                                                                    +++Globals+++                                                                     
|         1         |    9.75143     |        9.87224        |        100         |      100       |     10.255     |     10.255     |
                                                                +++Process Details+++                                                                 
|    consumer_0     |      100       |  12:29:44.694  |  12:29:54.412  |    0.07808     |    0.12330     |    0.09872     |    9.71772     |        9.87224        |
```

2. This PR with `python-snappy` installed

```
$ python scripts/perf_benchmark.py --num-processes 1 --num-events 100 --payload-bytes 50000000
                                                                    +++Globals+++                                                                     
|         1         |    5.78494     |       84.99988        |        100         |      100       |     17.286     |     17.286     |
                                                                +++Process Details+++                                                                 
|    consumer_0     |      100       |  12:31:04.519  |  12:31:10.278  |    0.09295     |    1.77285     |    0.85000     |    5.75875     |       84.99988        |

``` 

This PR almost cuts the `toatl_time` spent for the benchmark in half.

## How was it fixed?

1. We check if `python-snappy` is importable
2. If yes, then for every event we do:
  2.1 first pickle and then compress the bytes through snappy and then propagate
  2.2 when we receive we first decompress and then depickle into a Python object
3. If not, we send Python objects directly (which are pickled behind the scenes) as before

#### Cute Animal Picture


cburgdorf:
@pipermerriam I tried the same with `gzip.compress(...)` / `gzip.decompress(...)`  but it actually makes propagation times in this benchmark 3x worse than without expression and roughly 5x worse compared to snappy compression.

While gzip provides a very good compression rate it seems it's just too CPU intensive to do the job.

Using snappy compression if available seems to be a quick win.

pipermerriam:
This looks like a reasonably easy win.  I think that we should add an `extras_require` so that this can be installed with py-snappy in a single command **and** a CI test run that runs this with snappy installed (and ideally a test that does indeed verify that snappy is installed so that we don't get into a situation where we *think* this was being tested but it actually wasn't because the import was failing in CI).

And lastly, separate from this we *might* want to make a wrapper for `snappy` like `eth-hash` and `eth-keys` at some point which allows for the pure-python implementation or the fast one to be seemlessly swapped so that we can more easily negotiate flexibility in the system dependency in libraries that use them without the overhead of figuring out the system installs and things in every environment.

cburgdorf:
@pipermerriam I've added the jobs `py36-snappy-core`, `py37-snappy-core` and `snappy-benchmark`. In other words, we run all tests and the benchmark with and without snappy.

Btw, checking the benchmark output one can see that using snappy compression is just a tiny bit slower for very small events but gives a huge performance boost the bigger the payload gets.

lahja/endpoint.py
```diff
@@ -190,6 +205,20 @@ def start_serving_nowait(self,
             loop=self.event_loop
         )
 
+    def _compress_event(self, event: BaseEvent) -> Union[BaseEvent, bytes]:
+        if self.has_snappy_support:
```
pipermerriam: Wondering if we could gain a little more by removing the negotiation and do all of this in the `__init__`.  Rough pseudo-code sketch:

```python
def __init__(self, ...).
    if is_snappy_available(...):
        self.compress = cytoolz.compose(snappy.compress, pickle.dumps)
        self.decompress = cytoolz.compose(pickle.loads, snappy.decompress)
    else:
        self.compress = cytoolz.identity
        self.decompress = cytoolz.identity
```

cburgdorf: I tried it but I couldn't measure any notable difference in speed.

PR #36, opened 2019-03-19
Only send subscribed messages

lithp:
@cburgdorf Here's a prototype of a different approach to #34. This PR also only sends messages which the remote endpoint is listening for, it doesn't require any interface changes or cognitive overhead for the user, and as a bonus it dynamically changes the set of messages being listened for.

It switches on the type of the event so it doesn't support only subscribing to peer `NewTransaction` messages, for instance, but that could be added:
- `filter_predicate` parameters could be added to `subscribe()` and `stream()` and the callables sent to the remote `ConnectionFilter`s.
- the `PeerPoolMessageEvent` events could be broken up into multiple types, that would give us the best performance.

The code needs to be cleaned up a lot, so this is still a WIP, but what do you think about this approach?

### DIFF
--- a/lahja/endpoint.py
+++ b/lahja/endpoint.py
@@ -1,10 +1,15 @@
 import asyncio
 import functools
+import itertools
 import logging
+import multiprocessing
 from multiprocessing.managers import (  # type: ignore # Typeshed definition is lacking `BaseProxy`
     BaseManager,
     BaseProxy,
+    ConditionProxy,
+    RemoteError,
 )
+import multiprocessing.util
 import pathlib
 import pickle
 import threading
@@ -13,7 +18,6 @@ from typing import (  # noqa: F401
     AsyncGenerator,
     Callable,
     Dict,
-    Iterable,
     List,
     NamedTuple,
     Optional,
@@ -83,6 +87,9 @@ class EndpointConnector:
         # won't pick it up until some other task moves the loop forward
         loop.call_soon_threadsafe(self._endpoint._receiving_queue.put_nowait, item_and_config)
 
+    def subscribed_events(self) -> Set[Type[BaseEvent]]:
+        return self._endpoint.subscribed_events
+
 
 class ProxyEndpointConnector(BaseProxy):  # type: ignore # TypeShed is missing BaseProxy
     """
@@ -91,6 +98,49 @@ class ProxyEndpointConnector(BaseProxy):  # type: ignore # TypeShed is missing B
 
     def put_nowait(self, item_and_config: Tuple[BaseEvent, Optional[BroadcastConfig]]) -> None:
         self._callmethod('put_nowait', (item_and_config,))
+
+    def subscribed_events(self) -> Set[Type[BaseEvent]]:
+        return cast(Set[Type[BaseEvent]], self._callmethod('subscribed_events'))
+
+
+class ConnectorFilter:
+    """
+    Filter out messages the remote endpoint doesn't want to receive
+    """
+
+    def __init__(self,
+                 proxy: ProxyEndpointConnector,
+                 condition: ConditionProxy) -> None:
+        self.proxy = proxy
+        self.condition = condition
+        self.allowed_events: Set[Type[BaseEvent]] = proxy.subscribed_events()
+
+    def start(self) -> None:
+        threading.Thread(target=self._run, daemon=True).start()
+
+    def _run(self) -> None:
+        try:
+            while True:
+                with self.condition:
+                    self.update_subscription()
+                    self.condition.wait()
+        except (EOFError, BrokenPipeError, ConnectionResetError, RemoteError):
+            # Often the multiproxessing server shuts down before this thread does, leading
+            # to exceptions when we try to call methods on our ConditionProxy.
+            if not multiprocessing.util.is_exiting():
+                raise
+
+    def update_subscription(self) -> None:
+        self.allowed_events = self.proxy.subscribed_events()
+
+    def put_nowait(self, item_and_config: Tuple[BaseEvent, Optional[BroadcastConfig]]) -> None:
+        item, config = item_and_config
+        is_response = config is not None and config.filter_event_id
+        if type(item) in self.allowed_events or is_response:
+            self.proxy.put_nowait(item_and_config)
+
+
+Self = TypeVar('Self')
 
 
 class Endpoint:
@@ -119,7 +169,15 @@ class Endpoint:
         self._handler: Dict[Type[BaseEvent], List[Callable[[BaseEvent], Any]]] = {}
         self._queues: Dict[Type[BaseEvent], List[asyncio.Queue]] = {}
 
+        self._subscriptions_changed = multiprocessing.Condition()
+
         self._running = False
+
+    def __enter__(self: Self) -> Self:
+        return self
+
+    def __exit__(self, exc_type=None, exc_value=None, exc_tb=None) -> None:  # type: ignore
+        self.stop()
 
     @property
     def ipc_path(self) -> pathlib.Path:
@@ -153,6 +211,15 @@ class Endpoint:
                 self._has_snappy_support = False
 
         return self._has_snappy_support
+
+    @property
+    def subscribed_events(self) -> Set[Type[BaseEvent]]:
+        "Return the set of events this Endpoint is currently listening for"
+        return set(self._handler.keys()) | set(self._queues.keys())
+
+    def _notify_subscriptions_changed(self) -> None:
+        with self._subscriptions_changed:
+            self._subscriptions_changed.notify_all()
 
     def start_serving_nowait(self,
                              connection_config: ConnectionConfig,
@@ -248,13 +315,14 @@ class Endpoint:
             self._process_item(item, config)
 
     def _create_external_api(self, ipc_path: pathlib.Path) -> None:
-
+        condition = self._subscriptions_changed
         receiver = EndpointConnector(self)
 
         class ConnectorManager(BaseManager):
             pass
 
         ConnectorManager.register('get_connector', callable=lambda: receiver)  # type: ignore
+        ConnectorManager.register('Condition', callable=lambda: condition)  # type: ignore
 
         manager = ConnectorManager(address=str(ipc_path))  # type: ignore
         server = manager.get_server()   # type: ignore
@@ -309,10 +377,18 @@ class Endpoint:
             'get_connector',
             proxytype=ProxyEndpointConnector
         )
+        ConnectorManager.register('Condition', proxytype=ConditionProxy)  # type: ignore
 
         manager = ConnectorManager(address=str(endpoint.path))  # type: ignore
         manager.connect()
-        self._connected_endpoints[endpoint.name] = manager.get_connector()  # type: ignore
+
+        connector_filter = ConnectorFilter(
+            manager.get_connector(),  # type: ignore
+            manager.Condition(),  # type: ignore
+        )
+        connector_filter.start()
+
+        self._connected_endpoints[endpoint.name] = connector_filter
 
     def is_connected_to(self, endpoint_name: str) -> bool:
         return endpoint_name in self._connected_endpoints
@@ -430,8 +506,13 @@ class Endpoint:
         casted_handler = cast(Callable[[BaseEvent], Any], handler)
 
         self._handler[event_type].append(casted_handler)
+        self._notify_subscriptions_changed()
 
-        return Subscription(lambda: self._handler[event_type].remove(casted_handler))
+        def remove() -> None:
+            self._handler[event_type].remove(casted_handler)
+            self._notify_subscriptions_changed()
+
+        return Subscription(remove)
 
     TStreamEvent = TypeVar('TStreamEvent', bound=BaseEvent)
 
@@ -450,25 +531,25 @@ class Endpoint:
             self._queues[event_type] = []
 
         self._queues[event_type].append(queue)
-        i = None if num_events is None else 0
-        while True:
-            try:
-                yield await queue.get()
-            except GeneratorExit:
-                self._queues[event_type].remove(queue)
-                break
-            except asyncio.CancelledError:
-                self._queues[event_type].remove(queue)
-                break
-            else:
-                if i is None:
-                    continue
+        self._notify_subscriptions_changed()
 
-                i += 1
+        if num_events is None:
+            # loop forever
+            iterations = itertools.repeat(True)
+        else:
+            iterations = itertools.repeat(True, num_events)
 
-                if i >= cast(int, num_events):
-                    self._queues[event_type].remove(queue)
+        try:
+            for _ in iterations:
+                try:
+                    yield await queue.get()
+                except GeneratorExit:
                     break
+                except asyncio.CancelledError:
+                    break
+        finally:
+            self._queues[event_type].remove(queue)
+            self._notify_subscriptions_changed()
 
     TWaitForEvent = TypeVar('TWaitForEvent', bound=BaseEvent)
 
--- a/lahja/endpoint.py
+++ b/lahja/endpoint.py
@@ -133,10 +133,12 @@ class ConnectorFilter:
     def update_subscription(self) -> None:
         self.allowed_events = self.proxy.subscribed_events()
 
-    def put_nowait(self, item_and_config: Tuple[BaseEvent, Optional[BroadcastConfig]]) -> None:
+    def put_nowait(self,
+                   item_type: Type[BaseEvent],
+                   item_and_config: Tuple[BaseEvent, Optional[BroadcastConfig]]) -> None:
         item, config = item_and_config
         is_response = config is not None and config.filter_event_id
-        if type(item) in self.allowed_events or is_response:
+        if item_type in self.allowed_events or is_response:
             self.proxy.put_nowait(item_and_config)
 
 
@@ -450,7 +452,7 @@ class Endpoint:
             for name, connector in self._connected_endpoints.items():
                 allowed = (config is None) or config.allowed_to_receive(name)
                 if allowed:
-                    connector.put_nowait((compressed_item, config))
+                    connector.put_nowait(type(item), (compressed_item, config))
 
     TResponse = TypeVar('TResponse', bound=BaseEvent)
 
--- a/tests/core/test_filter_pred.py
+++ b/tests/core/test_filter_pred.py
@@ -1,0 +1,128 @@
+import asyncio
+from typing import (  # noqa: F401
+    List,
+)
+
+import pytest
+
+from conftest import (
+    generate_unique_name,
+)
+from lahja import (
+    BaseEvent,
+    BroadcastConfig,
+    ConnectionConfig,
+    Endpoint,
+)
+
+
+@pytest.mark.asyncio
+async def test_subscribed_events(event_loop: asyncio.AbstractEventLoop) -> None:
+    class TestSubscriptionEvent(BaseEvent):
+        pass
+
+    with Endpoint() as endpoint:
+        endpoint.subscribe(TestSubscriptionEvent, lambda event: None)
+
+        assert endpoint.subscribed_events == {TestSubscriptionEvent}
+
+        class TestStreamEvent(BaseEvent):
+            pass
+
+        async def stream() -> None:
+            async for _ in endpoint.stream(TestStreamEvent):  # noqa: F841
+                pass
+
+        future = asyncio.ensure_future(stream())
+        await asyncio.sleep(0)  # give stream a chance to be scheduled
+
+        assert endpoint.subscribed_events == {TestSubscriptionEvent, TestStreamEvent}
+
+        future.cancel()
+
+
+class TestEvent(BaseEvent):
+    # This must be importable in order for it to be pickleable
+    pass
+
+
+@pytest.mark.asyncio
+async def test_remote_subscribed_events(event_loop: asyncio.AbstractEventLoop) -> None:
+    with Endpoint() as local, Endpoint() as remote:
+        local.subscribe(TestEvent, lambda event: None)
+
+        assert local.subscribed_events == {TestEvent}
+
+        local_config = ConnectionConfig.from_name(generate_unique_name())
+        await local.start_serving(local_config, event_loop)
+
+        remote._loop = event_loop  # Endpoint assumes you'll serve before connecting
+        await remote.connect_to_endpoints(local_config)
+
+        proxy = remote._connected_endpoints[local_config.name]
+        assert proxy.proxy.subscribed_events() == {TestEvent}
+
+
+@pytest.mark.asyncio
+async def test_can_wait_for_changes(event_loop: asyncio.AbstractEventLoop) -> None:
+    with Endpoint() as local, Endpoint() as remote:
+        assert local.subscribed_events == set()
+
+        local_config = ConnectionConfig.from_name(generate_unique_name())
+        await local.start_serving(local_config, event_loop)
+
+        remote._loop = event_loop  # Endpoint assumes you'll serve before connecting
+        await remote.connect_to_endpoints(local_config)
+
+        proxy = remote._connected_endpoints[local_config.name]
+        assert proxy.proxy.subscribed_events() == set()
+
+        await asyncio.sleep(0.1)  # give the thread time to start
+
+        local.subscribe(TestEvent, lambda event: None)
+
+        await asyncio.sleep(0.1)  # give the thread time to update
+
+        assert proxy.proxy.subscribed_events() == {TestEvent}
+
+
+@pytest.mark.asyncio
+async def test_filtered_messages_not_sent(event_loop: asyncio.AbstractEventLoop) -> None:
+    "Test that messages we're not listening for are truly not sent"
+    class MockEndpoint(Endpoint):
+        def __init__(self) -> None:
+            super().__init__()
+            self.received_messages: List[BaseEvent] = []
+
+        def _process_item(self, item: BaseEvent, config: BroadcastConfig) -> None:
+            self.received_messages.append(item)
+
+    with MockEndpoint() as local, Endpoint() as remote:
+        local_config = ConnectionConfig.from_name(generate_unique_name())
+        await local.start_serving(local_config, event_loop)
+
+        remote_config = ConnectionConfig.from_name(generate_unique_name())
+        await remote.start_serving(remote_config, event_loop)
+        await remote.connect_to_endpoints(local_config)
+
+        assert len(local.received_messages) == 0
+
+        # We're not subscribed to the message so it better not be sent
+
+        remote.broadcast(TestEvent())
+
+        await asyncio.sleep(0.1)  # give the message time to propogate
+
+        assert len(local.received_messages) == 0
+
+        # Now, try subscribing to the message and see that it's sent
+
+        local.subscribe(TestEvent, lambda event: None)
+
+        await asyncio.sleep(0.1)  # give the thread time to update
+
+        remote.broadcast(TestEvent())
+
+        await asyncio.sleep(0.1)  # give the message time to propogate
+
+        assert len(local.received_messages) == 1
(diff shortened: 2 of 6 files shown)


lahja/endpoint.py
```diff
@@ -139,6 +183,15 @@ def logger(self) -> logging.Logger:
     def name(self) -> str:
         return self._name
 
+    @property
+    def subscribed_events(self) -> Set[Type[BaseEvent]]:
+        "returns the set of events this Endpoint is currently listening for"
```
cburgdorf: *Nitpick*: [We use imperative, present tense](https://github.com/ethereum/snake-charmers-tactical-manual/blob/master/documentation.md#grammar) (same as Python docs)

>Return the set of events this Endpoint is currently listening for

lahja/endpoint.py
```diff
@@ -420,15 +488,18 @@ def subscribe(self,
...
 
         self._queues[event_type].append(queue)
+        self._notify_subscriptions_changed()
         i = None if num_events is None else 0
         while True:
             try:
                 yield await queue.get()
             except GeneratorExit:
                 self._queues[event_type].remove(queue)
+                self._notify_subscriptions_changed()
```
cburgdorf: Maybe a small cleanup helper function grouping these two repeated lines would be justified now.

lahja/endpoint.py
```diff
@@ -90,6 +95,43 @@ class ProxyEndpointConnector(BaseProxy):  # type: ignore # TypeShed is missing B
     def put_nowait(self, item_and_config: Tuple[BaseEvent, Optional[BroadcastConfig]]) -> None:
         self._callmethod('put_nowait', (item_and_config,))
 
+    def subscribed_events(self) -> Set[Type[BaseEvent]]:
+        return self._callmethod('subscribed_events')
+
+
+class ConnectorFilter:
+    """
+    Filters out messages the remote endpoint doesn't want to receive
```
cburgdorf: Same here, imperative present tense:

>Filter out messages the remote endpoint doesn't want to receive

lahja/endpoint.py
```diff
@@ -90,6 +95,43 @@ class ProxyEndpointConnector(BaseProxy):  # type: ignore # TypeShed is missing B
...
+                self.condition.wait()
+            self.update_subscription()
+
+    def update_subscription(self):
+        self.allowed_events = self.proxy.subscribed_events()
+
+    def put_nowait(self, item_and_config: Tuple[BaseEvent, Optional[BroadcastConfig]]) -> None:
+        item, config = item_and_config
+        is_response = config is not None and config.filter_event_id
+        is_request = hasattr(item, '_id')
```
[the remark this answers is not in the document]
lithp: Yeah, I agree that using `_id` isn't ideal! In my defense, `BaseEvent` also assumes it :) Some of the tests do too, so if this were to change it would be caught pretty quickly.

https://github.com/ethereum/lahja/blob/e7928a8935fc77be332028b970fc6ae261970f87/lahja/misc.py#L42-L58

It looks like `filter_endpoint` doesn't work though, nothing enforces that it will be set. For example, it's not set in any of lahja's tests! I think I have a solution though, updates incoming.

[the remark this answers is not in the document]
cburgdorf: It's used in this test.

https://github.com/ethereum/lahja/blob/db081141b7184429ff60bd9e1673a1483773c2ee/tests/core/test_broadcast_config.py#L65

lahja/endpoint.py
```diff
@@ -489,28 +490,23 @@ def remove():
...
+        if num_events:
+            iterations = itertools.repeat(True, num_events)
+        else:
+            iterations = itertools.repeat(True)
 
-                if i >= cast(int, num_events):
-                    self._queues[event_type].remove(queue)
-                    self._notify_subscriptions_changed()
+        try:
+            for _ in iterations:
```
cburgdorf: Instead of the `itertools.repeat(True, num_events)` wouldn't something like `for i in range(num_events)` come more natural?

lithp: Yeah, I'm pretty frustrated by this code! But `range(None)` and even `itertools.repeat(True, None)` both throw exceptions, rather than iterating forever. And I could pull this out into a helper, but think that whatever name I could pick would more more confusing than just inlineing this code.

lithp: Looks like `islice` does the thing this needs! Just pushed a slightly cleaner commit.

lahja/endpoint.py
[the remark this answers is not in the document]
lithp: > It's used in this test.

Heh, oops, s/any/many/, good catch!

> Ok, curious what you come up with!

I spent some time thinking about this yesterday, and I can't find any obviously right solution.

`is_request` really shouldn't need to be there. If there's nothing on the other side listening for this request then there's no point in sending the request! A future PR could even notice that `filter_endpoint` was used and throw an exception rather than block the calling coro forever while it waits for a response which will never come.

However, removing `is_request` causes the tests to fail, and this is because of an annoying race condition: 

https://github.com/ethereum/lahja/blob/3a9a107ea6d7be893b0e2526838b95aed5f384d4/tests/core/test_basics.py#L17-L28

Here `endpoint.request()` blocks forever because this thread moves too quickly. There's no time for `notify_all()` to tell the endpoint about the new subscription or for the endpoint to react. This can be fixed by adding an `await asyncio.sleep(0.05)`, but that'd have to be added to many of the tests.

I don't think this will be a problem in the real world. I don't think we have anything which subscribes and expects to immediately receive messages, where a dropped initial message would cause problems. Since everything is in different processes trinity doesn't have the same idea of simultaneity that these lahja tests have.

There are some options for dealing with this:
1. Just add `asyncio.sleep()` to the tests. There's good precedent here, we often call `asyncio.sleep(0)` to allow other coroutines a chance to run.
2. Add a helper, `await_subscriptions_updated(endpoint, event_type)`, which blocks until the endpoint notices the subscription.
3. Somehow make `_notify_subscriptions_changed` synchronous. I couldn't find an easy way to do this, there's not really a way for `Server`s to send messages back to their clients.
4. Somehow make an `asyncio` version of `BaseManager` such that while everything is in the same process we don't need to use any extra threads, and `asyncio.sleep(0)` is enough to register a subscription.
5. Give our Endpoints the ability to inspect the `BaseManager` and notice that the remote Endpoint is running in the same thread as we are. When that's true `notify_subscriptions_changed` does something different which is synchronous, so we never have a race condition.

I don't like any of these, do you have any thoughts?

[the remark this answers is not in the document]
cburgdorf: >If there's nothing on the other side listening for this request then there's no point in sending the request!

Correct!

>I don't think this will be a problem in the real world. I don't think we have anything which subscribes and expects to immediately receive messages, where a dropped initial message would cause problems. Since everything is in different processes trinity doesn't have the same idea of simultaneity that these lahja tests have.

Yes, I agree, there shouldn't be a guarantee about which message will be the initial one receives after subscribing.

Solution 2. would be my favorite as well but I'm easy to please so I would even be ok with 1. :smile:

lahja/endpoint.py
pipermerriam: This block of code could use something like a big block comment at the top which explains what it is doing and why.  It's not clear to me what it's doing or why.

Also, `for _ in itertools.count()` will iterate forever...

pipermerriam: The code above doing the `islice` probably deserves a comment.  It wasn't clear to me on a quick read that it was also handling the possibility that `num_events` is `None` and thus needing it to do an infinite iteration in that case.

tests/core/test_filter_pred.py
```diff
@@ -0,0 +1,88 @@
...
+    future = asyncio.ensure_future(stream())
+    await asyncio.sleep(0)  # give stream a chance to be scheduled
+
+    assert endpoint.subscribed_events == {TestSubscriptionEvent, TestStreamEvent}
+
+    future.cancel()
+    endpoint.stop()
+
+
+class TestEvent(BaseEvent):
```
pipermerriam: Annoying but calling this `EventForTesting` will help prevent pytest from collecting it as a test.

tests/core/test_filter_pred.py
```diff
@@ -0,0 +1,88 @@
...
+    await local.start_serving(local_config, event_loop)
+
+    remote = Endpoint()
+    remote._loop = event_loop  # Endpoint assumes you'll serve before connecting
+    await remote.connect_to_endpoints(local_config)
+
+    proxy = remote._connected_endpoints[local_config.name]
+    assert proxy.proxy.subscribed_events() == {TestEvent}
+
+    remote.stop()
```
pipermerriam: This looks like something that should be doable via a context manager or in a `try/finally`.  What happens if we forget to explicitly stop these or the assertions fail and these statements don't get hit?

tests/core/test_filter_pred.py
pipermerriam: Kind of annoying, but if you name this `EventForTesting` then it pytest won't try to collect it as a test class.

lithp:
A quick update on this. The benchmarking test failed with an exception, and I've also seen it fail a different way when running locally. Here's a gist with both failures: https://gist.github.com/lithp/35f66be963b6cf68c1e3fedaa18b814d

The code runs just fine during normal execution, but multiprocessing's `atexit` handlers cause exceptions when trying to tear everything down.

A solution might be:
- Put a try-catch block somewhere and swallow exceptions which we know can only be caused by the program quitting
- The problem happens because we have long-lived `condition.wait()` calls which are still active when the process exits. A different architecture where endpoints directly told their remotes that their subscriptions had updated wouldn't have this problem. It would be a little more complicated though.

I'm looking at exactly why the failures occur during shutdown, hopefully there's a way to get it to shutdown cleanly without resorting to either of the above solutions.

lithp:
Today I tried going with a different approach: it might be easier to reason about Endpoints if they always communicate with each other directly. Instead of using a thread to long-poll Endpoints connect to the Endpoints which have connected to them and synchronously send messages to tell each other about changes to the message they're subscribed to. I'm still in the process of getting it working.

tests/core/test_filter_pred.py
lithp: I think we're save, pytest doesn't seem to be collecting this: 

```
(lahja) brian@rhythm:~/ef/lahja$ pytest --collect-only
========================================================================================================= test session starts =========================================================================================================
platform linux -- Python 3.7.0, pytest-4.0.2, py-1.8.0, pluggy-0.9.0 -- /home/brian/.local/share/virtualenvs/lahja-2m5rLbjb/bin/python3.7
cachedir: .pytest_cache
rootdir: /home/brian/ef/lahja, inifile: pytest.ini
plugins: xdist-1.25.0, forked-1.0.2, asyncio-0.9.0
collected 22 items                                                                                                                                                                                                                    
<Module 'tests/core/test_basics.py'>
  <Function 'test_request'>
  <Function 'test_request_can_get_cancelled'>
  <Function 'test_response_must_match'>
  <Function 'test_stream_with_break'>
  <Function 'test_stream_with_num_events'>
  <Function 'test_stream_can_get_cancelled'>
  <Function 'test_stream_cancels_when_parent_task_is_cancelled'>
  <Function 'test_wait_for'>
  <Function 'test_wait_for_can_get_cancelled'>
<Module 'tests/core/test_broadcast_config.py'>
  <Function 'test_broadcasts_to_all_endpoints'>
  <Function 'test_broadcasts_to_specific_endpoint'>
<Module 'tests/core/test_connect.py'>
  <Function 'test_can_not_connect_conflicting_names_blocking'>
  <Function 'test_can_not_connect_conflicting_names'>
  <Function 'test_rejects_duplicates_when_connecting_blocking'>
  <Function 'test_rejects_duplicates_when_connecting'>
  <Function 'test_rejects_duplicates_when_connecting_nowait'>
<Module 'tests/core/test_filter_pred.py'>
  <Function 'test_subscribed_events'>
  <Function 'test_remote_subscribed_events'>
  <Function 'test_can_wait_for_changes'>
<Module 'tests/core/test_import.py'>
  <Function 'test_import'>
<Module 'tests/core/test_internal.py'>
  <Function 'test_internal_propagation'>
<Module 'tests/core/test_stop.py'>
  <Function 'test_can_stop'>
```

tests/core/test_filter_pred.py
lithp: Just wrapped it into a context manager.

lahja/endpoint.py
lithp: Just pushed a version which is a more verbose but also easier to read.

lithp:
Okay, I think this is ready for review again. I fixed all the atexit exceptions and squashed history and added a test that messages which are filtered out really aren't being sent.

I found it pretty hard to get this working, and I think a big piece is that multiprocessing doesn't play nicely with asyncio. A bigger change might be to change how we're using multiprocessing or replace it with something like zeromq which has an asyncio binding, I think that would make this code a lot easier to work with.

If you don't mind, I think `filter_predicate` can be added as part of a different PR, this one seems to be working! And that's a further performance optimization that can be added in the future.

lithp:
Well... it looks like `perf_benchmark` now deadlocks. I'll look into it tomorrow :(

lahja/endpoint.py
```diff
@@ -92,6 +99,49 @@ class ProxyEndpointConnector(BaseProxy):  # type: ignore # TypeShed is missing B
...
+        try:
+            while True:
+                with self.condition:
+                    self.update_subscription()
+                    self.condition.wait()
+        except (EOFError, BrokenPipeError, ConnectionResetError, RemoteError):
+            # Often the multiproxessing server shuts down before this thread does, leading
+            # to exceptions when we try to call methods on our ConditionProxy.
+            if not multiprocessing.util.is_exiting():
+                raise
```
cburgdorf: Clever

pipermerriam:
> or we could implement something with `asyncio.Streams`

This seems like a good hack-day project for while we're here together in June.  I suspect a minimal implementation should be pretty straight forward to do.

lithp:
Okay, all the benchmarking now works every time. I know you've already approved it twice but this was a big enough change to `perf_benchmark.py` that I don't want to just merge without giving you a chance to review it one last time.

Feel free to merge it yourself if you want to build off this! And if I don't hear back I'll merge it sometime next week.

cburgdorf:
Awesome! Looks good! The only thing I want to double check is if it work in Trinity or if any adjustments would be needed. I don't think so but I think it's worth checking. I'm on mobile now but can check it later today or tomorrow.

cburgdorf:
Ok, it looks like it doesn't work as a drop-in replacement in Trinity. It breaks the system where the endpoints connect to each other. There might be some changes in Trinity needed to adjust to this PR but I haven't looked into it yet.

lithp:
Ah, that's frustrating, and also a little embarrassing; I should have checked myself. What did it break in trinity?

cburgdorf:
>What did it break in trinity?

The endpoints aren't subscribing each other which then causes things like the discovery not answering the peer pool etc. I guess what happens is that the endpoints send `EndpointConnected` to `main` before the `main` process has subscribed.

lithp:
It looks like Ctrl-c also got uglier:

```
    INFO  04-09 19:16:49               trinity  DB process (pid=19580) terminated
Exception in thread Thread-39:
Traceback (most recent call last):
  File "/home/brian/.local/share/virtualenvs/trinity-mNs_Su9C/lib/python3.7/site-packages/lahja/endpoint.py", line 126, in _run
    self.condition.wait()
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/managers.py", line 1038, in wait
    return self._callmethod('wait', (timeout,))
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/managers.py", line 796, in _callmethod
    kind, result = conn.recv()
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/connection.py", line 250, in recv
    buf = self._recv_bytes()
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/connection.py", line 407, in _recv_bytes
    buf = self._recv(4)
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/connection.py", line 383, in _recv
    raise EOFError
EOFError

During handling of the above exception, another exception occurred:

Traceback (most recent call last):
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/threading.py", line 917, in _bootstrap_inner
    self.run()
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/threading.py", line 865, in run
    self._target(*self._args, **self._kwargs)
  File "/home/brian/.local/share/virtualenvs/trinity-mNs_Su9C/lib/python3.7/site-packages/lahja/endpoint.py", line 126, in _run
    self.condition.wait()
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/managers.py", line 1032, in __exit__
    return self._callmethod('release')
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/managers.py", line 795, in _callmethod
    conn.send((self._id, methodname, args, kwds))
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/connection.py", line 206, in send
    self._send_bytes(_ForkingPickler.dumps(obj))
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/connection.py", line 404, in _send_bytes
    self._send(header + buf)
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/connection.py", line 368, in _send
    n = write(self._handle, buf)
BrokenPipeError: [Errno 32] Broken pipe
```

lithp:
I think I should admit failure here, I thought this approach would be simpler and the user interface it exposes it certainly easier to use, but it was obvious how `filter_predicate` worked and with that approach we wouldn't have had to worry about these kinds of race conditions or messy exits.

in this PR the sending endpoints wait for any updates by long-polling into each of the endpoints they connect to. Because of how multiprocessing works this required a thread, and because Python doesn't support interrupting threads there's no easy way I can think of to make shutdowns definitely cleaner. *Also* because of the long-polling there's no easy way to block until other threads have updated their subscriptions, so we get race-conditions.

I think an easy fix to the race conditions is to add a flag which lets you say "I know the other side will want this message" so it skips the filter. That would only be a stop-gap until this is replaced with a better architecture. I'm currently looking into why this fails on Trinity, hopefully that or a similar fix is enough.

I also have a branch which gets rid of the threads and communicates via endpoints connecting back to the endpoints which connect to them, then sending messages the usual way. I think I'll take a look at getting that working too.

cburgdorf:
>I think I should admit failure here, I thought this approach would be simpler and the user interface it exposes it certainly easier to use, but it was obvious how `filter_predicate` worked and with that approach we wouldn't have had to worry about these kinds of race conditions or messy exits.

The `fiter_predicate` solution is certainly simpler implementation wise it has worked well for me so far but it is in fact very appealing not having to worry about manually setting up filters for efficiency!

>I think an easy fix to the race conditions is to add a flag which lets you say "I know the other side will want this message" so it skips the filter.

So, you mean the sender would use something like a `force` flag in the `BroadcastConfig` to bypass the filter? This would then only be used for certain event types where the initial event is important such as `EndpointConnected` and maybe `ShutdownRequested`.

>I also have a branch which gets rid of the threads and communicates via endpoints connecting back to the endpoints which connect to them, then sending messages the usual way. I think I'll take a look at getting that working too.

Nice! But I guess, the main problem remains that we have no certainty at the time of sending an event whether we do already know about the receivers possible subscription or not. I guess what we would need for that is that  `connect_to_endpoints`  would not only wait until we have established the connection to the remote endpoint but also send an initial info about its possible subscriptions to our events.

If you feel that the complexity passes a certain threshold and takes away too much of your time that you rather spend on the sync effort, we can also fall back to the `filter_predicate` solution and push the transparent solution down the road. That said, no pressure from me as I'm on vacation until 24th anyway. I just wanted to throw this in as an option, so you know you shouldn't feel obligated to finish the solution if the effort gets too high after all.

lithp:
> So, you mean the sender would use something like a force flag in the BroadcastConfig to bypass the filter? This would then only be used for certain event types where the initial event is important such as EndpointConnected and maybe ShutdownRequested.

Yeah, exactly!

> Nice! But I guess, the main problem remains that we have no certainty at the time of sending an event whether we do already know about the receivers possible subscription or not. I guess what we would need for that is that connect_to_endpoints would not only wait until we have established the connection to the remote endpoint but also send an initial info about its possible subscriptions to our events.

Yep, the idea is to make that part synchronous. `connect_to_endpoints` shouldn't return until you can safely send messages.

> If you feel that the complexity passes a certain threshold and takes away too much of your time that you rather spend on the sync effort, we can also fall back to the filter_predicate solution and push the transparent solution down the road. That said, no pressure from me as I'm on vacation until 24th anyway. I just wanted to throw this in as an option, so you know you shouldn't feel obligated to finish the solution if the effort gets too high after all.

Thanks for the offer! And enjoy your vacation :) I want to give this another try but I'll keep that in mind and switch back to state sync if this takes much longer.

Today I spent more time looking into why this fails with Trinity. I'm still not sure what's happening but I've discovered some things:
- `_connect_receiving_queue` is a coroutine, but `handle_new_endpoints` calls `connect_to_endpoints_blocking`, which blocks the main thread which it tries to connect.
- `_connect_receiving_queue` is also in a precarious spot. It doesn't catch any exceptions, so any exception which a `handler` raises causes it to stop processing messages. Even worse, nothing is printed! I think that's because the coro is passed into `gather()`, `asyncio` plans on returning the exception along with the `gather()` call.
- For some reason, when `main` tries to connect to `bjson-rpc-api` it receives `[Errno 111] Connection refused`.
- When `main` tries to connect to `discovery`, it receives `[Errno 104] Connection reset by peer`

That last exception prevents networking from being told about the discovery service (the `AvailableEndpointsUpdated` message is never sent) and therefore it's requests to the discovery service are dropped.

lithp:
> For some reason, when main tries to connect to bjson-rpc-api it receives [Errno 111] Connection refused.

After spending some quality time with strace, it appears that this happens because the plugin segfaults soon after it starts listening. Here's the output from a core dump:


```gdb
Program terminated with signal SIGSEGV, Segmentation fault.
#0  __new_sem_wait_fast (definitive_result=1, sem=<optimized out>) at sem_waitcommon.c:135
135	sem_waitcommon.c: No such file or directory.
[Current thread is 1 (Thread 0x7f82a88a9080 (LWP 4921))]
(gdb) bt
#0  __new_sem_wait_fast (definitive_result=1, sem=<optimized out>) at sem_waitcommon.c:135
#1  __new_sem_trywait (sem=0x7fe8718de000) at sem_wait.c:82
#2  0x00007f82a603b183 in semlock_acquire (self=0x7f82a3777d18, args=<optimized out>, kwds=<optimized out>)
    at /home/brian/.pyenv/sources/3.7.0/Python-3.7.0/Modules/_multiprocessing/semaphore.c:309
#3  0x000000000043beca in _PyMethodDef_RawFastCallKeywords (method=0x7f82a623c560 <semlock_methods+64>, self=self@entry=0x7f82a3777d18, args=args@entry=0x7f829c2f99e8, 
    nargs=nargs@entry=0, kwnames=kwnames@entry=0x0) at Objects/call.c:690
#4  0x00000000005bc349 in _PyMethodDescr_FastCallKeywords (descrobj=0x7f82a6447048, args=args@entry=0x7f829c2f99e0, nargs=nargs@entry=1, kwnames=kwnames@entry=0x0)
    at Objects/descrobject.c:288
#5  0x000000000042acb9 in call_function (kwnames=0x0, oparg=1, pp_stack=<synthetic pointer>) at Python/ceval.c:4563
(gdb) py-bt
Traceback (most recent call first):
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/synchronize.py", line 96, in __enter__
    return self._semlock.__enter__()
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/synchronize.py", line 231, in __enter__
    return self._lock.__enter__()
  File "/home/brian/ef/lahja/lahja/endpoint.py", line 261, in _notify_subscriptions_changed
    with self._subscriptions_changed:
  File "/home/brian/ef/lahja/lahja/endpoint.py", line 571, in subscribe
    self._notify_subscriptions_changed()
  File "/home/brian/ef/trinity/trinity/endpoint.py", line 54, in auto_connect_new_announced_endpoints
    self.subscribe(AvailableEndpointsUpdated, self.connect_to_other_endpoints)
  File "/home/brian/ef/trinity/trinity/extensibility/plugin.py", line 318, in _prepare_start
    self.event_bus.auto_connect_new_announced_endpoints()
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/process.py", line 99, in run
    self._target(*self._args, **self._kwargs)
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/process.py", line 297, in _bootstrap
    self.run()
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/spawn.py", line 118, in _main
    return self._bootstrap()
  File "/home/brian/.pyenv/versions/3.7.0/lib/python3.7/multiprocessing/spawn.py", line 105, in spawn_main
    exitcode = _main(fd)
  File "<string>", line 1, in <module>
```

lithp:
The same segfault happens to `PeerDiscoveryPlugin`, which would explain why `main` can't connect back to it.

lithp:
After spending more time with strace I've figured out why the plugins are segfaulting. They segfault while calling: `self._notify_subscriptions_changed()`, which uses a semaphore internally. The problem is that this semaphore was created in the `main` process and the memory it was mapped to does not exist in the child processes.

The process is failing on a dereference:
```gdb
(gdb) disas
Dump of assembler code for function __new_sem_trywait:
=> 0x00007f0922dea840 <+0>:	mov    (%rdi),%rax
   0x00007f0922dea843 <+3>:	test   %eax,%eax
   0x00007f0922dea845 <+5>:	je     0x7f0922dea858 <__new_sem_trywait+24>
   0x00007f0922dea847 <+7>:	lea    -0x1(%rax),%rdx
...
(gdb) info registers
...
rdi            0x7fe8718da000	140636314247168
```

That address is not a part of any of the memory segments, however it does show up in the main process' strace:

```gdb
lstat("/dev/shm/DsUT5s", 0x7ffcf6436c80) = -1 ENOENT (No such file or directory)
openat(AT_FDCWD, "/dev/shm/DsUT5s", O_RDWR|O_CREAT|O_EXCL, 0600) = 22
write(22, "\1\0\0\0\0\0\0\0\200\0\0\0\350\177\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0\0", 32) = 32
mmap(NULL, 32, PROT_READ|PROT_WRITE, MAP_SHARED, 22, 0) = 0x7fe8718da000
link("/dev/shm/DsUT5s", "/dev/shm/sem.mp-8zv2nf9_") = 0
fstat(22, {st_mode=S_IFREG|0600, st_size=32, ...}) = 0
unlink("/dev/shm/DsUT5s")               = 0
close(22)                               = 0
unlink("/dev/shm/sem.mp-8zv2nf9_")      = 0
```

The child process is created by these lines:

https://github.com/ethereum/trinity/blob/413882806fe5031e790fa0be11ef1e037ed06886/trinity/extensibility/plugin.py#L284-L294

Under the covers, `Process.start()` is using a `clone()` followed by an `execve()` which wipes away *everything* and starts afresh. I can't find the code which restores state and makes `self` work, but the child process spends a lot of time reading data the parent gave it and memmapping, so it seems like multiprocessing somehow (pickles and?) sends all the state to the child process. It does not, however, send the semaphores over, so the child crashes when it tries to notify.

This would be fixed by not creating an event bus for the Plugin until the new process has already been started.

lithp:
Yeah, that does the trick. I've opened [a PR against trinity](https://github.com/ethereum/trinity/pull/500) which changes it to not send `Endpoint`s over process boundaries. Now everything appears to be working!

lithp:
Closing in favor of #42. #42 is based on asyncio which should cause fewer headaches in the long run.

PR #39, opened 2019-04-21
Replace multiprocessing with asyncio

### LINKED ISSUE #37: Pure asyncio?
I've been somewhat passively paying attention to the work that both @cburgdorf and @lithp are doing.  I don't claim to know it in detail, but my impression is that there has been a reasonable amount of difficulty that is in some way related to `multiprocessing`

I want to pose something that's been talked about with a small-ish working code example.

Below is a simplistic concept for an event bus that does direct connection and allows transport of any pickle-able object.  Does it make sense for us to go this route at this point?  Is the somewhat heavy non-asyncio friendly machinery of the `multiprocessing` library getting in the way?

```python
import asyncio
import logging
from pathlib import Path
import pickle


class Endpoint:
    def __init__(self, path: Path) -> None:
        self.path = path
        self.running = asyncio.Event()
        self.finished = asyncio.Event()

    async def start_server(self) -> None:
        await asyncio.start_unix_server(
            self.accept_connection,
            path=str(self.path),
        )
        # set the running event
        self.running.set()

    async def accept_connection(self, reader, writer) -> None:
        self.reader, self.writer = reader, writer
        logging.info('INCOMING CONNECTION: %s', self.path)
        self.server = asyncio.ensure_future(self.listen())

    @classmethod
    async def connect(cls, path: Path) -> None:
        reader, writer = await asyncio.open_unix_connection(
            str(path),
        )
        client = cls(path)
        client.reader = reader
        client.writer = writer
        client.server = asyncio.ensure_future(client.listen())
        logging.info('OUTBOUND CONNECTION: %s', path)
        return client

    async def listen(self) -> None:
        try:
            while not self.reader.at_eof():
                raw_size = await self.reader.readexactly(4)
                size = int.from_bytes(raw_size, 'little')
                message = await self.reader.readexactly(size)
                obj = pickle.loads(message)

                logging.info('RECEIVED OBJECT: %s', obj)
        finally:
            self.finished.set()

    async def broadcast(self, obj):
        message = pickle.dumps(obj)
        size = len(message)
        self.writer.write(size.to_bytes(4, 'little'))
        self.writer.write(message)
        await self.writer.drain()
        logging.info('SENT OBJECT: %s', obj)

    async def close(self):
        if self.writer.can_write_eof():
            self.writer.write_eof()
            await self.writer.drain()
        self.server.cancel()
        self.writer.close()
        # set the finished event
        self.finished.set()


# tests.py
from pathlib import Path
from typing import NamedTuple

import pytest

from purebus import Endpoint


IPC_PATH = Path('./test.ipc')


class User(NamedTuple):
    name: str
    age: int


@pytest.fixture
async def server():
    endpoint = Endpoint(IPC_PATH)
    await endpoint.start_server()
    yield endpoint
    await endpoint.close()


@pytest.mark.asyncio
async def test_connection(server):
    await server.running.wait()

    client = await Endpoint.connect(server.path)
    await client.broadcast(b'ping')
    await client.broadcast(1)
    await client.broadcast(User('piper', 35))
    await client.close()
```

Here's the logging output from running this.

```bash
$ pytest tests.py --log-cli-level=INFO
============================================================ test session starts =============================================================
platform linux -- Python 3.6.5, pytest-3.6.2, py-1.7.0, pluggy-0.6.0 -- /home/piper/python-environments/evm/bin/python
cachedir: .pytest_cache
rootdir: /home/piper/projects/pure-bus, inifile:
plugins: xdist-1.18.1, cov-2.5.1, asyncio-0.9.0, asyncio-network-simulator-0.1.0a2, hypothesis-3.69.5
collected 1 item

tests.py::test_connection
--------------------------------------------------------------- live log call ----------------------------------------------------------------
purebus.py                  69 INFO     OUTBOUND CONNECTION: test.ipc
purebus.py                  90 INFO     SENT OBJECT: b'ping'
purebus.py                  90 INFO     SENT OBJECT: 1
purebus.py                  90 INFO     SENT OBJECT: User(name='piper', age=35)
PASSED                                                                                                                                 [100%]
------------------------------------------------------------- live log teardown --------------------------------------------------------------
purebus.py                  57 INFO     INCOMING CONNECTION: test.ipc
```

lithp:
Closes #37

Builds on top of #43

This comes with some interface changes:
- connect_to_endpoints_blocking no longer exists, there's no way to
  start a server without using await
- it's important that endpoint.stop() is called when the endpoint is no
  longer needed. If it is not called a lot of coros will be orphaned. (Endpoint is now a context manager which makes this easier)
- broadcast() is now an awaitable. Previously backpressure would stall the main thread, now it only stalls the coroutine which is trying to send data.
- broadcast_nowait() schedules a message to be broadcast and returns immediately. Notably, the messages is not yet sent when this method returns! It'll be sent the next time the event loop loops.

Future work:
- [ ] The code might be simpler if `CompressedBroadcast` is pulled out from `Broadcast`.

Some open questions:
- the existing code calls `asyncio.gather` but never awaits the result. I'm surprised that this isn't a problem.
- Is it okay that I removed `connect_to_endpoints_blocking`? `trinity` uses it and I'm not sure how big of a refactor it would take to make that not true.
- Why did I have to change so many tests? For some reason, once I made these changes `pytest` started giving very messy output when I ran the tests, because all sorts of coros were being orphaned: `Task was destroyed but it is pending!`, `RuntimeError: Event loop is closed`. To the best of my knowledge all of these coros were *already* being orphaned, and I don't understand what my changes did to expose the problem.

#### Cute Animal Picture

cutest-bunny-rabbits-01


pipermerriam:
I think it's very reasonable for this to introduce breaking API changes.  We should just make sure they are documented and that we do proper version bumping. (I think breaking changes are allowed in the `0.N -> 0.N+1` in semver????)

pipermerriam:
Looks like benchmarks still time out.   Have you gotten them to run?  If so have you compared performance?

pipermerriam:
Oh, I think a footgun with this approach (mostly because we used multiprocessing a lot before) is that these don't seem to complain when passed across process boundaries (pickled -> unpickled) but I think that the `StreamReader` and `StreamWriter` objects stop working when this happens.  I don't know exactly why but it's something i saw in my experiment.

lithp:
> Looks like benchmarks still time out. Have you gotten them to run? If so have you compared performance?

Wow, I had run them but I hadn't looked at the numbers, the results are (mostly) encouraging!

On my computer:

| Flags  | master | pure-asyncio |
| ------------- | ------------- | ----- |
| num-processes=10 num-events=10000  | 17s  | 6.11s |
| num-processes=3 num-events=10000  | 4.07s  | 1.74s | 
| num-processes=3 num-events=1000 payload-bytes=1M | 5 s | **crashes** |

I would have expected this version to be slightly slower, I'm excited that it's such a big performance win! 

The crash is quite exciting:

```python
Task exception was never retrieved
future: <Task finished coro=<RemoteEndpoint.send_message() done, defined at /home/brian/ef/lahja/lahja/endpoint.py:76> exception=AssertionError()>
Traceback (most recent call last):
  File "/home/brian/ef/lahja/lahja/endpoint.py", line 77, in send_message
    await _write_message(self.writer, message)
  File "/home/brian/ef/lahja/lahja/endpoint.py", line 95, in _write_message
    await writer.drain()
  File "/home/brian/.pyenv/versions/3.6.7/lib/python3.6/asyncio/streams.py", line 339, in drain
    yield from self._protocol._drain_helper()
  File "/home/brian/.pyenv/versions/3.6.7/lib/python3.6/asyncio/streams.py", line 214, in _drain_helper
    assert waiter is None or waiter.cancelled()
AssertionError
```

I guess I should look into why we crash once we start to see any backpressure, that sounds important :sweat_smile:

lithp:
> Oh, I think a footgun with this approach (mostly because we used multiprocessing a lot before) is that these don't seem to complain when passed across process boundaries (pickled -> unpickled) but I think that the StreamReader and StreamWriter objects stop working when this happens. I don't know exactly why but it's something i saw in my experiment.

Ah, that's frustrating. Yeah, I don't think the Server should pickle well, I'll look into generating an error message when you try to pickle.

lithp:
I took a before and after of the performance: https://gist.github.com/lithp/ad6ea2c2a3815295679ed680d788f091

Now, if only I knew why it failed to run on circle ci, it works on my computer every time :cry:

pipermerriam:
That performance win is so nice.

lahja/endpoint.py
```diff
@@ -202,9 +227,33 @@ def start_serving_nowait(self,
...
+    async def _handle_client(self, remote: RemoteEndpoint) -> None:
+        while self._running and remote.more_to_read:
+            message = await remote.read_message()
+
+            if isinstance(message, Broadcast):
+                await self._receiving_queue.put((message.event, message.config))
+            else:
+                self.logger.warning(
+                    f'[{self.ipc_path}] received unexpected message: {message}'
+                )
```
pipermerriam: It would be nice if this could cleanup it's own `coro` from `self._child_coros` so that set doesn't end up with a bunch of finished/done/cancelled coros hanging around.  Not entirely clear how we'd go about doing that but ideas off top of my head are:

- use a mapping of `RemoteEndpoint -> coro` so this function can know which one to reference.
- use [`Future.add_done_callback`](https://docs.python.org/3/library/asyncio-future.html#asyncio.Future.add_done_callback) to manage the cleanup...
- maybe use a `weakref.WeakKeyDictionary` so that we can just track the `RemoteEndpoint` instance and once we ditch it the future cleans up automatically.

lahja/endpoint.py
```diff
@@ -61,36 +64,52 @@ def from_name(cls, name: str, base_path: Optional[pathlib.Path]=None) -> 'Connec
...
+            # https://bugs.python.org/issue29930
+            await self.writer.drain()
 
-    def put_nowait(self, item_and_config: Tuple[BaseEvent, Optional[BroadcastConfig]]) -> None:
-        self._callmethod('put_nowait', (item_and_config,))
+    async def read_message(self) -> Broadcast:
+        return await _read_message(self.reader)
+
+    @property
+    def more_to_read(self) -> bool:
```
pipermerriam: Alternate name ideas.

- `has_more_to_read` 
- `is_reader_done`
- `is_reader_at_eof`  (most literal/explicit)
- `can_read`  (meh on this one)

lahja/tools/benchmark/process.py
```diff
@@ -56,79 +61,85 @@ def start(self) -> None:
...
+        loop = asyncio.get_event_loop()
+        loop.run_until_complete(DriverProcess.worker(config))
+
+    @staticmethod
+    async def worker(config: DriverProcessConfig) -> None:
+        with Endpoint() as event_bus:
+            await event_bus.start_serving(ConnectionConfig.from_name(DRIVER_ENDPOINT))
+            await event_bus.connect_to_endpoints(*config.connected_endpoints)
+
+            payload = b'\x00' * config.payload_bytes
```
pipermerriam: Not needed for this PR but it might be good to try and use representative data.

- text from a book that's in the creative commons
- data from the [`faker`](https://faker.readthedocs.io/en/master/providers.html)

lahja/tools/benchmark/process.py
```diff
@@ -56,79 +61,85 @@ def start(self) -> None:
...
-        loop = asyncio.get_event_loop()
-        event_bus = Endpoint()
-        event_bus.start_serving_nowait(ConnectionConfig.from_name(name))
-        event_bus.connect_to_endpoints_blocking(
-            ConnectionConfig.from_name(REPORTER_ENDPOINT)
-        )
+    def launch(name: str, num_events: int, ready_event: Event) -> None:
         # UNCOMMENT FOR DEBUGGING
         # logger = multiprocessing.log_to_stderr()
         # logger.setLevel(logging.INFO)
```
pipermerriam: Ideally we move this to the test running level so we can control it with a CLI flag from pytest.

pipermerriam: https://github.com/ethereum/lahja/issues/41

lahja/tools/benchmark/process.py
pipermerriam: https://github.com/ethereum/lahja/issues/40

lahja/endpoint.py
```diff
@@ -61,36 +64,52 @@ def from_name(cls, name: str, base_path: Optional[pathlib.Path]=None) -> 'Connec
...
+        return not self.reader.at_eof()
+
+
+async def _read_message(reader: StreamReader) -> Broadcast:
+    raw_size = await reader.readexactly(4)
```
cburgdorf: Should we put the `4` into a constant (e.g. `SIZE_MARKER_LENGTH`)?

lahja/endpoint.py
```diff
@@ -61,36 +64,52 @@ def from_name(cls, name: str, base_path: Optional[pathlib.Path]=None) -> 'Connec
...
+    obj = pickle.loads(message)
+    assert isinstance(obj, Broadcast)  # maybe this should be a user-friendly error?
+    return obj
+
+
+async def _write_message(writer: StreamWriter, message: Broadcast) -> None:
+    assert isinstance(message, Broadcast)
+    pickled = pickle.dumps(message)
+    size = len(pickled)
+    writer.write(size.to_bytes(4, 'little'))
```
cburgdorf: So this would refer to `SIZE_MARKER_LENGTH` (for the lack of a better name) as well

lahja/endpoint.py
cburgdorf: Looking at the call sites, I find `has_more_to_read` the most pleasant to use.

lahja/endpoint.py
```diff
@@ -247,38 +296,25 @@ def _throw_if_already_connected(self, *endpoints: ConnectionConfig) -> None:
...
+        raise NotImplementedError()
 
     async def connect_to_endpoints(self, *endpoints: ConnectionConfig) -> None:
         """
         Connect to the given endpoints and await until all connections are established.
         """
         self._throw_if_already_connected(*endpoints)
+        loop = None
+        if self._loop:
+            loop = self.event_loop
```
cburgdorf: Maybe this is still in the "make it run" phase but I just wanted to point out that this looks suspicious to me. *Locally* falling back to `loop=None` if `self._loop` isn't set doesn't strike me as right. Especially because any attempt to read from `self.event_loop` would throw at this time.

lahja/endpoint.py
```diff
@@ -355,8 +381,21 @@ def stop(self) -> None:
...
+        return self
+
+    def __exit__(self,
+                 exc_type: Optional[Type[BaseException]],
+                 exc_value: Optional[BaseException],
+                 exc_tb: Optional[TracebackType]) -> None:
+        self.stop()
 
-    def broadcast(self, item: BaseEvent, config: Optional[BroadcastConfig] = None) -> None:
+    async def broadcast(self, item: BaseEvent, config: Optional[BroadcastConfig] = None) -> None:
```
cburgdorf: I believe making this `async` will result in `async` creep and maybe difficult to resolve situations within Trinity. Nothing specific comes to my mind but I *think*  we use `broadcast` from quite a bunch of synchronous methods which may be hard or awkward to turn into `async`.

Do we need the backpressure for this to work? Or could this be an opt-in thing resulting in another method `broadcast_nowait` maybe? I quite like that `broadcast` has worked in a fire-and-forget manner so far. What happens if we broadcast an event to many consumers and one of them (e.g. a third party plugin) consumes them too slowly because of some blocking code. Doesn't that slow down the event propagation for all other consumers as well?

cburgdorf: On a second thought, I guess the `broadcast_nowait` could simply be implemented as a `asyncio.ensure_future(...)` wrapping the call like you did further down in the benchmark. But if there's a legit need for it I think adding this as a method such as `broadcast_nowait` on the endpoint would make a lot of sense.

pipermerriam: I'm also :+1: on a `_nowait` option as I think @cburgdorf is correct about the difficulties we'll find in trinity.

lahja/endpoint.py
lithp: Sorry, I don't understand the complaint :disappointed: The logic defaults to using the current loop (which definitely exists, because this is an async method), that lets this function be called before `start_serving`. However, if our loop exists we use that instead.

I agree this isn't very pretty! but I think it's correct?

cburgdorf: >The logic defaults to using the current loop (which definitely exists, because this is an async method), that lets this function be called before `start_serving`.

But isn't that potentially problematic? What if it uses the default loop but then `start_serving` is called with a different one (e.g. a special one in another thread).

lahja/endpoint.py
lithp: I was worried about this one too but I think making `broadcast` async is the right thing to do. `broadcast` sends data into a socket and that socket could genuinely back up.

>  Or could this be an opt-in thing resulting in another method broadcast_nowait maybe? 

If it's really necessary I guess we can add this, but this is a scary method! There are two ways to do it:
- `broadcast` before this PR would just write into the socket, this blocked the event loop whenever the socket filled up. (and also blocked the event loop until the remote process had confirmed receipt, I think this is why performance has improved)
- `ensure_future` is how I had originally written this PR, but it failed when I ran: `perf_benchmark.py --num-processes 3 --num-events 10000 --payload-bytes 1000000`. This test fired off a bunch of futures, each of them holding a MB of data, each of them waiting to be able to write to the socket. For reasons I don't quite understand the process ended up consuming ~26GB of memory and swapping wildly as it tried to send. This is an extreme example, but it convinced me that it's much safer for `broadcast` to block the caller.

> What happens if we broadcast an event to many consumers and one of them (e.g. a third party plugin) consumes them too slowly because of some blocking code. Doesn't that slow down the event propagation for all other consumers as well?

I think it usually won't. The `InboundConnection` reads messages out and throws them into the unbounded `_receiving_queue`, so a slow plugin would only cause memory bloat in it's own process. An exception to this would be a plugin which manages to block the event loop. That would eventually fill the socket's buffer and start blocking broadcast.

It might be possible to detect this situation and emit a warning? I don't have a good answer for you, but I lean towards poor performance being better than unbounded memory growth.

lahja/endpoint.py
lithp: Okay, good point, this is a little dangerous. It'd be nice to have some way of enforcing that Endpoint is always called from the same loop.

lahja/endpoint.py
cburgdorf: >If it's really necessary I guess we can add this, but this is a scary method!

I think it's ok to discourage usage of `broadcast_nowait` in the docstring but I still prefer to have it when the alternative is to force uses to use `asyncio.ensure_future(...) or even `cast(None, asyncio.ensure_future(...))` when used in a `lambda`. Especially, because `broadcast_nowait` can also set `loop=self._loop` to ensure we broadcast on the right loop and not on the default loop.

lithp:
Just rebased this branch onto #43

lahja/endpoint.py
lithp: I ended up going with `reader_not_at_eof`. I agree that `has_more_to_read` sounds nicest but I don't think I understand `at_eof` well enough to try to abstract it away behind a name, this is the name which makes the fewest promises.

Happy to discuss if you think something else is better!

lahja/tools/benchmark/process.py
lithp: Yeah, definitely! This is a reason why at one point benchmarking would hang (on the massive messages) but snappy-benchmark worked, the snappy messages compress down to nothing :smile:

lahja/endpoint.py
lithp: Coming back to this, I sat down to try to change it but I'm not sure what to do here. I think it's valuable to be able to connect to remote endpoints without first serving, so either of them could be called first. There doesn't seem to be any good reason to move an Endpoint between event loops though.

 It seems like the safest thing is to, whenever a method is called:
- If `self._loop` is not already set, set it
And either one of:
- If `self._loop` is already set, use that loop
- If `self._loop` is already set, raise an exception if it is not equal to `asyncio.get_event_loop()`

What do you think?

lahja/endpoint.py
cburgdorf: Sounds good! :+1:

lahja/endpoint.py
cburgdorf: >If `self._loop` is already set, raise an exception if it is not equal to `asyncio.get_event_loop()`

That would make the most sense to me!

lithp:
@cburgdorf This one is also ready for you!

lahja/endpoint.py
lithp: Okay, I wrote a `check_event_loop` decorator which guards functions which use the loop and disallows dangerous calls.

lahja/endpoint.py
```diff
@@ -69,6 +68,10 @@ class Broadcast(NamedTuple):
     config: Optional[BroadcastConfig]
 
 
+class ReaderAtEof(Exception):
```
cburgdorf: Maybe that should also move to `exceptions.py` and inherit from `LahjaException`?

lahja/endpoint.py
```diff
@@ -162,6 +170,23 @@ def event_loop(self) -> asyncio.AbstractEventLoop:
 
         return self._loop
 
+    def check_event_loop(func: TFunc) -> TFunc:  # type: ignore
```
cburgdorf: Do I read that right that this means the `Endpoint` can not run with a different event loop than the default event loop? This may get into trouble with the Trinity tests where we create an event loop explicitly.

https://github.com/ethereum/trinity/blob/b7eaa8a195aecbacfd982d03adfde976ded6298f/tests/conftest.py#L105-L143

lahja/endpoint.py
```diff
@@ -155,36 +209,69 @@ def has_snappy_support(self) -> bool:
...
-        asyncio.ensure_future(self._connect_internal_queue(), loop=self.event_loop)
+        self._internal_loop_running = asyncio.Event()
+        self._receiving_loop_running = asyncio.Event()
+        self._server_running = asyncio.Event()
+        self._internal_queue = asyncio.Queue()
+        self._receiving_queue = asyncio.Queue()
+
+        # Using `gather` (over e.g. `wait` or plain `ensure_future`) ensures that the inner futures
+        # are automatically cancelled as soon as the parent task is cancelled
+        asyncio.gather(
```
cburgdorf: Why are we back to using `asyncio.gather` here (was just recently dropped https://github.com/ethereum/lahja/commit/43e479d4159afd99d70de36bac7902a4414cfb0c)

lahja/tools/benchmark/process.py
```diff
@@ -112,39 +122,41 @@ def launch(name: str, num_events: int) -> None:
...
+                timeout=5
+            )
+            try:
+                async for event in events:
+                    stats.add(RawMeasureEntry(
+                        sent_at=event.sent_at,
+                        received_at=time.time()
+                    ))
+            except asyncio.TimeoutError:
+                print('Some messages from the consumer were dropped')
```
cburgdorf: Answering to your comment https://github.com/ethereum/lahja/pull/43#discussion_r280908368 here.

>I'm starting to think that it's a bad idea to assume every message will be received, the sending process might start sending before the receiving process starts listening.

I think that our benchmark should ensure that we start sending messages only after the other process listens (which was the case so far) and that our benchmark should fail if messages get dropped. I believe accepting dropped messages in the benchmark could be dangerous because we may not notice a bug that would cause dropped messages for no plausible reason.

lahja/endpoint.py
```diff
@@ -133,6 +170,23 @@ def event_loop(self) -> asyncio.AbstractEventLoop:
 
         return self._loop
 
+    def check_event_loop(func: TFunc) -> TFunc:  # type: ignore
+        """
+        All Endpoint methods must be called from the same event loop.
+        """
+        @functools.wraps(func)
```
cburgdorf: I think this decorator might accidentally remove type information from any method that it is applied on. There is a `mypy` setting that would catch this that we currently don't have enabled for this project.

```
--disallow-any-decorated  Disallow functions that have Any in their
                            signature after decorator transformation
```

cburgdorf: I double checked this and it turns out, that this decorator does in fact preserve the type information. Also, we **do run** mypy in strict mode in this project. So, carry on :)

lahja/endpoint.py
cburgdorf: Ok, looking at your Trinity PR, it looks like we get around manually setting event loops. I'm ok going down that way :+1:

lahja/tools/benchmark/process.py
cburgdorf: Looks like, [this is the place](https://github.com/ethereum/lahja/pull/39/files#diff-d27046448bb81af24db4f2419116d9fcR69) that ensures the order in which the actions of the benchmark start is correct so save us from dropping events.

I'm much in favor of not sweeping dropped events under the rug then.

pipermerriam:
> The main issue I have is about sweeping dropped events under the rug in the benchmark. I think this is dangerous and I'd rather ensure the order in which things fire is well defined (which I think it is?) 

How about a middleground where we consume an infinite stream of events that can be sequentially ordered (like using a counter number) and we just benchmark a section of that stream and enforce proper ordering in the part that we consume.

- producer begins sending events `1, 2, 3, 4, ....., 10, 11, 12`
- consumer comes online and begins consuming either a fixed number of events or for a fixed amount of time: `3, 4, 5, 6, 7, 8` and then finishes, validating that it got a contiguous stream of events.
- producer is shutdown.

Could something like this work?

lahja/endpoint.py
lithp: Yeah, the change to trinity to allow this was surprisingly small!

It's still possible to use lahja on a non-default event loop. Lahja uses the event loop which is active the first time one of its async methods are called. The code is simpler but this drops the ability to tell lahja to use an event loop which has not yet been entered.

lahja/endpoint.py
lithp: Oops, good catch! I accidentally re-introduced it during a rebase.

lahja/endpoint.py
lithp: Heh, it's cool that `functools.wraps` does the right thing. I think that even if it didn't the outer signature (`def check_event_loop(func: TFunc) -> TFunc`) tells mypy that we're not changing the signature of any of the functions this wraps.

lithp:
>  The main issue I have is about sweeping dropped events under the rug in the benchmark. I think this is dangerous and I'd rather ensure the order in which things fire is well defined (which I think it is?) and let things blow up in case we ever experience dropped events. Wouldn't be good if at some point we miss events because of some obscure bug and it goes unnoticed. Sure, we also have tests in place but I see the benchmark as a second line of defense that might expose issues that the tests don't see.

Yeah, I see your point and I agree with you, allowing events to be dropped is sub-optimal and has a high chance of hiding bugs. My reason for adding this was that I've spent an insane amount of time trying to get `perf_benchmark.py` to not deadlock whenever I make a change to lahja. I think that I've had to do this because of an impedance mismatch. Lahja provides very loose coupling, "fire and forget". This benchmark, however assumes that all messages are delivered. My idea was that by changing the behavior this benchmark expects to be the behavior lahja actually provides I would fix these deadlocks once and for all.

This weekend I thought a little about going the other direction, what if lahja was changed to provide what this benchmark expects? A naive solution would be to remember all messages ever sent and received. When you connect to a new endpoint you send all the previously generated messages. When you subscribe to an event you replay all received messages which match the subscription. This is roughly the interface which kafka exposes; by writing all the messages to a file (which would be amazing for debugging) lahja could do this without consuming large amounts of memory.

> How about a middleground where we consume an infinite stream of events that can be sequentially ordered (like using a counter number) and we just benchmark a section of that stream and enforce proper ordering in the part that we consume.

I quite like this idea! This is the best of both worlds, I'll give it a shot.

lithp:
Addressed your comments @cburgdorf. And @pipermerriam's suggestion worked very well. This is a big enough change I wanted to give you one more chance to review it before it ships!

lahja/tools/benchmark/process.py
pipermerriam: Again, throwing solutions at problems I haven't really taken time to digest.  Using `multiprocessing.Event` in the processes which drive these seems like exactly the tool you're looking for...(don't broadcast until the receiving process has indicated it's ready by setting the event)

lithp: > Using multiprocessing.Event in the processes which drive these seems like exactly the tool you're looking for...

Yeah, totally! That's actually what I was doing in this PR before the last change, which instead follows your suggestion of having the driver keep sending messages until the consumers have received enough.

lithp:
> So, I think it might be a useful lahja feature but I don't see an immediate need for it so it kinda falls into my personal YAGNI bucket at least for now :)

Oh yeah, mine too :) Just wanted to bring it up as an alternative to think about next time lahja goes through changes.

PR #42, opened 2019-04-24
[WIP] (asyncio) Only send subscribed messages

lithp:
Super WIP, opened to see what CircleCI's tests say

A replacement for #36 built on top of #39.

TODO:
- [ ] Make subscribe() a regular call again

## What was wrong?

Issue #

## How was it fixed?

Summary of approach.

#### Cute Animal Picture


lahja/endpoint.py
```diff
@@ -61,36 +65,177 @@ def from_name(cls, name: str, base_path: Optional[pathlib.Path]=None) -> 'Connec
...
+Msg = Union[Broadcast, SubscriptionsUpdated, SubscriptionsAck]
+
+
+class Connection:
+    def __init__(self, reader: StreamReader, writer: StreamWriter) -> None:
+        self.writer = writer
+        self.reader = reader
+        self._drain_lock = asyncio.Lock()
+
+    @staticmethod
```
pipermerriam: This looks like it should be a `classmethod` and return an instance of `cls`

lahja/endpoint.py
```diff
@@ -61,36 +65,177 @@ def from_name(cls, name: str, base_path: Optional[pathlib.Path]=None) -> 'Connec
...
+        self.conn = conn
+        self.new_msg_func = new_msg_func
+
+        self.logger = logging.getLogger('lahja.endpoint.InboundConnection')
+
+        self._notify_lock = asyncio.Lock()
+        self._received_response = asyncio.Condition()
+
+    async def run(self) -> None:
+        try:
```
pipermerriam: I think this should be inverted, with the try/except inside the `while` loop such that the `except` block uses `break` to escape the loop.

lahja/endpoint.py
```diff
@@ -61,36 +65,177 @@ def from_name(cls, name: str, base_path: Optional[pathlib.Path]=None) -> 'Connec
...
+            while self.conn.more_to_read:
+                message = await self.conn.read_message()
+
+                if isinstance(message, Broadcast):
+                    self.new_msg_func(message)
+                elif isinstance(message, SubscriptionsAck):
+                    async with self._received_response:
+                        self._received_response.notify_all()
+                else:
+                    self.logger.error(f'received unexpected message: {message}')
```
pipermerriam: Might be good to include the `type(message)` here since it may not be clear what `message` is when represented as a string.

lahja/endpoint.py
```diff
@@ -61,36 +65,177 @@ def from_name(cls, name: str, base_path: Optional[pathlib.Path]=None) -> 'Connec
...
+        endpoint has acknowledged the new subscription set. If `block` is False then this
+        function will return immediately after the send finishes.
+        """
+        async with self._notify_lock:
+            # The lock ensures only one coroutine can notify this endpoint at any one time
+            # and that no replies are accidentally received by the wrong coroutines.
+            async with self._received_response:
+                await self.conn.send_message(SubscriptionsUpdated(subscriptions, block))
+                if block:
+                    await self._received_response.wait()
```
pipermerriam: I kind of understand this but only because I've spent some brain-space in the past trying to understand the use case for `asyncio.Condition`.  Can you maybe create a large-ish comment that tries to lay out exactly what is happening here and why.  This interplay between synchronization primatives seems appropriate but also hard for a newcomer to understand.

lahja/endpoint.py
```diff
@@ -61,36 +65,177 @@ def from_name(cls, name: str, base_path: Optional[pathlib.Path]=None) -> 'Connec
...
+    endpoint occasionally sends message "backwards" to the endpoint which connected to it.
+    Those messages (SubscriptionsUpdated) specify which kinds of messages the remote
+    Endpoint is subscribed to. All other messages should be dropped.
+    """
+    def __init__(self, name: str, conn: Connection) -> None:
+        self.conn = conn
+        self.name = name
+        self.subscribed_messages: Set[Type[BaseEvent]] = set()
+
+        self.logger = logging.getLogger('lahja.endpoint.OutboundConnection')
```
pipermerriam: This can be defined at the class level.

lahja/endpoint.py
```diff
@@ -61,36 +65,177 @@ def from_name(cls, name: str, base_path: Optional[pathlib.Path]=None) -> 'Connec
...
+        self._received_subscription = asyncio.Condition()
+
+    async def run(self) -> None:
+        try:
```
pipermerriam: Same here with reducing the scope of the `try` block to only contain the part that can throw the exception.

lahja/endpoint.py
```diff
@@ -61,36 +65,177 @@ def from_name(cls, name: str, base_path: Optional[pathlib.Path]=None) -> 'Connec
...
+                async with self._received_subscription:
+                    self._received_subscription.notify_all()
+                if message.response_expected:
+                    await self.send_message(SubscriptionsAck())
+        except asyncio.IncompleteReadError:
+            # the remote endpoint has disconnected
+            pass
+
+    def can_send_item(self, item: BaseEvent, config: Optional[BroadcastConfig]) -> bool:
+        is_response = config is not None and config.filter_event_id is not None
```
pipermerriam: Since there is an early exit condition for `passes_config`, maybe optimize this a little by doing that check before you compute `is_response` or `passes_filter` since they are not needed in the early exit case.

lahja/endpoint.py
```diff
@@ -109,17 +254,23 @@ class Endpoint:
     _internal_queue: asyncio.Queue
     _internal_loop_running: asyncio.Event
 
-    _loop: asyncio.AbstractEventLoop
+    _server_running: asyncio.Event
```
pipermerriam: This seems prudent to have as a public API since it is likely a common thing to want to wait until the endpoint is active...

lahja/endpoint.py
```diff
@@ -368,13 +554,14 @@ def broadcast(self, item: BaseEvent, config: Optional[BroadcastConfig] = None) -
...
-            compressed_item = self._compress_event(item)
-            for name, connector in self._connected_endpoints.items():
-                allowed = (config is None) or config.allowed_to_receive(name)
-                if allowed:
-                    connector.put_nowait((compressed_item, config))
+            return
+
+        # Broadcast to every connected Endpoint that is allowed to receive the event
+        compressed_item = self._compress_event(item)
+        for name, remote in list(self._outbound_connections.items()):
```
pipermerriam: *nitpick*.  In another place you use `copy`.  Here you cast to a `list`.  Can we use a common mechanism for this (like maybe `tuple()` because I love them so much more than lists :smile: )

lahja/tools/benchmark/process.py
```diff
@@ -92,36 +96,37 @@ def start(self) -> None:
...
+        with Endpoint() as event_bus:
+            await event_bus.start_serving(ConnectionConfig.from_name(name))
+            await event_bus.connect_to_endpoints(ConnectionConfig.from_name(REPORTER_ENDPOINT))
+
+            reporter = event_bus._outbound_connections[REPORTER_ENDPOINT]
+            await reporter.wait_until_subscribed_to(TotalRecordedEvent)
+
+            counter = itertools.count(1)
+            stats = LocalStatistic()
+            async for event in event_bus.stream(PerfMeasureEvent):
```
pipermerriam: Is there a reason this doesn't this use the `num_events` API that the `stream` method implements (instead of the current thing which uses an `itertools.count()` to exit the loop?  Then you could move the broadcasting of the `TotalRecordedEvents` into an `else` clause (or just after this for loop's block)

tests/core/test_broadcast_config.py
```diff
@@ -51,12 +55,12 @@
 
     tracker = Tracker()
 
-    endpoint1.subscribe(
+    await endpoint1.subscribe(
```
cburgdorf: This falls into the same category as the `broadcast` that became *async*. We uses this API in a bunch of places that would need to turn into `async` and I actually quite like that synchronous code could use `subscribe` to consume events :)

lithp: Yeah! I should come back and look at `subscribe`, my idea was that I wanted to use an architecture which got rid of these race conditions once and for all. So `subscribe` blocks until all the inbound connections have confirmed that they're only sending the requested messages. This was supposed to reduce the number of `asyncio.sleep()` I added to the tests. It didn't get rid of very many `asyncio.sleep()` calls though, and I later added the more useful `wait_until_subscribed_to`, so I'll look into simplifying the code and making `subscribe` a regular call again.

cburgdorf:
Going to review this once the other two PRs landed and this got rebased.

lithp:
Closing in favor of #55

lahja/tools/benchmark/process.py
lithp: Good catch! Did this in a different PR

PR #43, opened 2019-04-25
Make Endpoint.broadcast() async

lithp:
## What was wrong?

While working on #39 I attempted to make `Endpoint.broadcast()` async. That PR uses asyncio to communicate with other Endpoints, so any caller attempting to `broadcast()` must either be a coroutine which is okay with being blocked or call `asyncio.ensure_future`.

This PR makes some interface changes:
- `Endpoint.broadcast()` is now async
- `Endpoint.broadcast_nowait()` now exists and does what `Endpoint.broadcast()` used to do.
- `start_serving_nowait()` no longer exists. Callers must use `Endpoint.start_serving()`.
- `start_serving()` no longer accepts an event loop, it uses the current event loop.

I don't believe that this PR meaningfully changes behavior, but Christoph pointed out that #39, the reason for this PR, will block calling coroutines if other processes manage to block their own event loops. I think this is okay, because the other alternatives I see are:
- drop messages to remotes which are taking too long to process messages
- allow the local queue of messages to be sent to grow unboundedly

And both of those sound worse to me than bad performance.

I've also opened a PR against trinity which uses this branch: https://github.com/ethereum/trinity/pull/538

#### Cute Animal Picture

elephant

### DIFF
--- a/lahja/tools/benchmark/process.py
+++ b/lahja/tools/benchmark/process.py
@@ -1,9 +1,9 @@
 import asyncio
-import itertools
 import logging
 import multiprocessing
 import time
 from typing import (  # noqa: F401
+    AsyncGenerator,
     NamedTuple,
     Optional,
     Tuple,
@@ -101,6 +101,17 @@ class ConsumerProcess:
         loop.run_until_complete(ConsumerProcess.worker(name, num_events))
 
     @staticmethod
+    async def gen_with_timeout(generator: AsyncGenerator, timeout: int) -> AsyncGenerator:
+        try:
+            while True:
+                # the asyncio.TimeoutError is the caller's responsibility
+                future = asyncio.ensure_future(generator.__anext__())
+                future.add_done_callback(lambda x: None)
+                yield await asyncio.wait_for(future, timeout=timeout)
+        except StopAsyncIteration:
+            pass
+
+    @staticmethod
     async def worker(name: str, num_events: int) -> None:
         event_bus = Endpoint()
         await event_bus.start_serving(ConnectionConfig.from_name(name))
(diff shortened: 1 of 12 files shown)


tests/core/test_basics.py
```diff
@@ -17,11 +20,11 @@
 async def test_request(endpoint: Endpoint) -> None:
     endpoint.subscribe(
         DummyRequestPair,
-        lambda ev: endpoint.broadcast(
+        lambda ev: cast(None, asyncio.ensure_future(endpoint.broadcast(
```
cburgdorf: I really think we should add a `broadcast_nowait` that does this because:

- the `asyncio.ensure_future` becomes repetitive and even more involved if it also needs a `cast`
- in a `broadcast_nowait` we can make sure that `asyncio_ensure_future` puts the `broadcast` call on `loop=self._loop` which might be important for situations where the endpoint does not run on the default loop.

lithp: Okay, I'm not opposed to adding a `broadcast_nowait` :+1: 

> in a broadcast_nowait we can make sure that asyncio_ensure_future puts the broadcast call on loop=self._loop which might be important for situations where the endpoint does not run on the default loop.

What's a situation where you would have two loops in the same process? That sounds dangerous!

lithp:
Okay, I agree that sometimes you actually want `broadcast_nowait` so it should at least be available, even if it's discouraged. And I agree that it makes sense to wait to merge this until everything is fleshed out. I still like the idea of having a couple small PRs so I'll leave this hanging around and build off of it.

tests/core/test_basics.py
cburgdorf: I found it weird and scary, too when I first encountered it but we even had this within Trinity before the JSON-RPC server moved into an isolated plugin. Here is a link to [old Trinity code that did this](https://github.com/ethereum/trinity/blob/7b1fec6249652df01535ac900c190b0b538d2830/trinity/nodes/base.py#L143-L154).

There's also a comment on it that tries to explain the rational. I'm happy we got rid of it but there may be valid reasons for people to use that pattern for certain scenarios.

cburgdorf:
>And I agree that it makes sense to wait to merge this until everything is fleshed out. I still like the idea of having a couple small PRs so I'll leave this hanging around and build off of it.

Totally! :+1:

lahja/tools/benchmark/process.py
```diff
@@ -108,21 +120,25 @@ def launch(name: str, num_events: int) -> None:
...
-                event_bus.stop()
-                break
+        try:
+            async for event in events:
+                stats.add(RawMeasureEntry(
+                    sent_at=event.sent_at,
+                    received_at=time.time()
+                ))
+        except asyncio.TimeoutError:
+            print('Some messages from the consumer were dropped')
```
cburgdorf: Isn't a dropped event something that should fail the whole benchmark?

lithp: So, yes and no. After spending a lot of time trying to get this benchmark to work after each change I make to lahja I'm starting to think that it's a bad idea to assume every message will be received, the sending process might start sending before the receiving process starts listening.

This change prevents `perf_benchmark.py` from hanging forever if some messages are dropped. I think that in a vacuum dropped messages might be a reason to fail the benchmark, but #39 starts the processes in the right order so that they won't be dropped, so I hope that this PR can be merged as is, given that it'll be immediately replaced with #39.

lithp:
@cburgdorf This is ready for review! Let me know what you think.

lahja/tools/benchmark/process.py
cburgdorf: >so I hope that this PR can be merged as is, given that it'll be immediately replaced with #39.

I'm ok to merge it as is but as noted in #39, I believe we should then remove this again so that the benchmark would fail when messages get dropped.

PR #49, opened 2019-05-09
Standardize endpoint api

pipermerriam:
## What was wrong?

The current implementation of `Endpoint` doesn't have a firm definition of what the core API that an alternate endpoint would need to implement would be.  In order for us to implement a `TrioEndpoint` we need this definition to be firm.

## How was it fixed?

Defined an `ABC` base class which cements the core API for an endpoint.  This **should** be very close the the minimal necessary API that will work across different async frameworks and not require custom code to naively work with different endpoint implementations using the same base async code for most use cases.

#### Cute Animal Picture

d0b1e1064631d35efc22d0d6459e8181--miniture-cows-miniature-goats


pipermerriam:
- extract snappy change to stand-alone
- remove backend stuff from benchmarks
- extract reverse broadcast

lahja/base.py
```diff
@@ -0,0 +1,169 @@
...
+        raise NotImplementedError('Endpoints cannot be pickled')
+
+    async def wait_for(self, event_type: Type[TWaitForEvent]) -> TWaitForEvent:
+        """
+        Wait for a single instance of an event that matches the specified event type.
+        """
+        agen = self.stream(event_type, num_events=1)
+        event = await agen.asend(None)
+        await agen.aclose()
+        return event
```
pipermerriam: Wanted to specifically point out the change to how this method was implemented.

pipermerriam:
@cburgdorf this is ready for review

lahja/base.py
```diff
@@ -0,0 +1,169 @@
...
+        """
+        Return whether this endpoint is connected to another endpoint with the given name.
+        """
+        pass
+
+    #
+    # Event API
+    #
+    @abstractmethod
+    async def broadcast(self, item: BaseEvent, config: Optional[BroadcastConfig] = None) -> None:
```
cburgdorf: This should probably also add an abstract method for `def broadcast_nowait(self, ...)` no?

lahja/base.py
cburgdorf: I like how that is now implemented :+1:

lahja/base.py
pipermerriam: I'm currenty attempting to approach this as `_nowait` APIs being allowed in individual implementations but not part of the core required API.  While I know how to do it in `trio`, it is intentionally not easy and thus I'm trying to avoid those kind of side-effect APIs that toss something into the background and forget about it.

cburgdorf: Oh you are absolutely right. I forgot for a moment that `trio` is all about getting rid of these "wild" background taskes :) :+1:

PR #51, opened 2019-05-09
Simplify snappy support check

pipermerriam:
## What was wrong?

Snappy support check should be re-usable.  

## How was it fixed?

Moved it to a private namespace and made it easy to import without having to do error checking every place we want to check.

#### Cute Animal Picture

Cute-Goats-Photo-15

### DIFF
--- a/lahja/_snappy.py
+++ b/lahja/_snappy.py
@@ -1,0 +1,7 @@
+is_snappy_available: bool
+
+try:
+    import snappy  # noqa: F401
+    is_snappy_available = True
+except ModuleNotFoundError:
+    is_snappy_available = False
--- a/lahja/_snappy.py
+++ b/lahja/_snappy.py
@@ -1,7 +1,8 @@
-is_snappy_available: bool
 
-try:
-    import snappy  # noqa: F401
-    is_snappy_available = True
-except ModuleNotFoundError:
-    is_snappy_available = False
+
+def check_has_snappy_support() -> bool:
+    try:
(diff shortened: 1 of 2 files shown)


lahja/_snappy.py
```diff
@@ -0,0 +1,7 @@
+is_snappy_available: bool
+
+try:
```
cburgdorf: whats the reason for not making this a function that just returns `True` or `False`?

lahja/endpoint.py
```diff
@@ -198,12 +198,8 @@ def name(self) -> str:
...
     def has_snappy_support(self) -> bool:
         if self._has_snappy_support is None:
-            try:
-                import snappy  # noqa: F401
-                self._has_snappy_support = True
-            except ModuleNotFoundError:
-                self._has_snappy_support = False
-
+            from lahja._snappy import is_snappy_available
+            self._has_snappy_support = is_snappy_available
```
cburgdorf: Then this would become `self._has_snappy_support = has_snappy_support()`. There's probably something clever that I'm missing but from the top of my head I don't see the benefit of that construct.

pipermerriam: I think I'm :+1: with deciding at import time since this isn't something that should change over the lifecycle of the python process

cburgdorf: Ah, so now I understand the reason for that other approach. But yeah, I think assuming snappy support is available at the time when the process starts is a reasonable assumption to make.

PR #53, opened 2019-05-09
Use black code formatting

pipermerriam:
## What was wrong?

Code style..... not always the same.... can be distracting sometimes...

## How was it fixed?

don't think about it and use `black`

#### Cute Animal Picture

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -27,7 +27,8 @@ lint:
 	tox -elint
 
 lint-roll:
-	isort --recursive lahja tests scripts
+	isort --recursive lahja scripts tests
+	black lahja scripts tests
 	$(MAKE) lint
 
 test:
--- a/lahja/__init__.py
+++ b/lahja/__init__.py
@@ -1,10 +1,5 @@
-from .endpoint import (  # noqa: F401
-    Endpoint,
-)
-from .exceptions import (  # noqa: F401
-    ConnectionAttemptRejected,
-    UnexpectedResponse,
-)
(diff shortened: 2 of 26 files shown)


pipermerriam:
PR opened to gauge whether we're ready for this here in a smaller library.

ralexstokes:
i would support trying this out for a bit in a more isolated context at first

cburgdorf:
I'm :+1: as well. But maybe we wait until #42 was rebased or even wait some more days if there's currently more open work in flux. I'd assume rebasing changes on top of  a `black` reformatting might not be fun. It's not the end of the world but the black formatting is cheap and can easily be made after the dust has settled.

pipermerriam:
@ralexstokes am I correct interpreting that to mean *"yes, I support trying it out here in the `lahja` codebase where it is smaller and more isolated than it was in `trinity`"*

ralexstokes:
@pipermerriam yep, 👍  :) i realize that was more ambiguous than i intended now that i'm re-reading it...

cburgdorf:
@pipermerriam Unless you have some other open work that would be send in merge conflict hell by this change, I think we can bring this up to date and land it. I'd like to cut a new Lahja release this week and I think black could already be included there.

pipermerriam:
Nothing blocking this other than me rebasing/cleaning.  Let me do that real quick.

pipermerriam:
@cburgdorf do you want to :eyes: this before merge?

PR #54, opened 2019-05-09
Remove type hints from tests

### LINKED ISSUE #52: Remove type checking from tests
## What was wrong?

Currently the `./tests` files have type hints.  This was done to catch a typing bug where `Endpoint.stream` was returning `Any` types.

However, the overhead of typing our tests isn't something we want to continue with.

## How can it be fixed?

Remove all type hints from under `./tests` and create a new file similar to [this one in `eth-utils`](https://github.com/ethereum/eth-utils/blob/83df44277d87900d18e57ff91e2f307cf33549f8/tests/mypy_typing_of_functional_utils.py) and add a line in CI that checks the type hints for this file.

pipermerriam:
fixes #52 

## What was wrong?

`trio` specifies `mypy` as a dependency and it requires a newer version than `lahja` uses.

## How was it fixed?

Updated `mypy` to latest and fixed type hints.  Also went ahead and addressed #52 

#### Cute Animal Picture

2eabd7eb42d2b7e4baced9e41b698168

### DIFF
--- a/tests/mypy_typing_test.py
+++ b/tests/mypy_typing_test.py
@@ -1,4 +1,8 @@
-from typing import AsyncIterable, Type
+from typing import (
+    AsyncIterable,
+    Type,
+)
+
 from lahja import (
     BaseEvent,
     BaseRequestResponseEvent,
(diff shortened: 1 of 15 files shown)


tests/mypy_typing_test.py
```diff
@@ -0,0 +1,29 @@
...
+    return await endpoint.wait_for(Event)
+
+
+async def verify_stream_type(endpoint: Endpoint) -> AsyncIterable[Event]:
+    async for event in endpoint.stream(Event):
+        yield event
+
+
+async def verify_request_response_type(endpoint: Endpoint) -> Event:
+    return await endpoint.request(RequestEvent())
```
cburgdorf: This is missing coverage of the `subscribe` API. Something like:

```
async def verify_subscribe_handler_type(endpoint: Endpoint) -> None:
    def handler(event: Event) -> None:
        pass
     endpoint.subscribe(Event, handler)
```

PR #57, opened 2019-05-14
Set event loop in Endpoint.run if not already set

pipermerriam:
## What was wrong?

The `Endpoint.run` method should set the event loop on the endpoint upon entry.

## How was it fixed?

Check if the `self._loop` is unset and set it if it is.

#### Cute Animal Picture

5951153_f520

### DIFF
--- a/lahja/endpoint.py
+++ b/lahja/endpoint.py
@@ -376,7 +376,9 @@ class Endpoint(BaseEndpoint):
         self.ipc_path.unlink()
 
     @asynccontextmanager  # type: ignore
-    async def run(self) -> AsyncIterable['Endpoint']:
+    async def run(self) -> AsyncIterable['Endpoint']:  # type: ignore
+        if not self._loop:
+            self._loop = asyncio.get_event_loop()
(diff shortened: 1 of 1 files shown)


lahja/endpoint.py
```diff
@@ -376,7 +376,9 @@ def stop(self) -> None:
         self.ipc_path.unlink()
 
     @asynccontextmanager  # type: ignore
-    async def run(self) -> AsyncIterable['Endpoint']:
+    async def run(self) -> AsyncIterable['Endpoint']:  # type: ignore
```
cburgdorf: Shouldn't this use the `check_event_loop` decorator that all the other methods also use?

pipermerriam:
I don't know what just happened.  I was looking at a page with this having an approval and tests passing but it must have been old.  Gimme a minute to sort this out

lahja/endpoint.py
pipermerriam: Something about the `asynccontextmanager` didn't play well with with it.

PR #59, opened 2019-05-14
setup.py linting cleanup

pipermerriam:
## What was wrong?

The `setup.py` file format caused linting errors in my editor.

## How was it fixed?

Changed from using line broken list addition to use `list.extend` to build the `[dev]` install requirements.

#### Cute Animal Picture

Goat+Farming+in+Nigeria

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -34,12 +34,11 @@ extras_require = {
     ]
 }
 
-extras_require['dev'] = (
-    extras_require['dev'] +
-    extras_require['test'] +
-    extras_require['lint'] +
-    extras_require['doc']
-)
+
+extras_require['dev'].extend(extras_require['test'])
+extras_require['dev'].extend(extras_require['lint'])
+extras_require['dev'].extend(extras_require['doc'])
+
 
 setup(
     name='lahja',


setup.py
```diff
@@ -34,12 +34,11 @@
...
-    extras_require['dev'] +
-    extras_require['test'] +
-    extras_require['lint'] +
-    extras_require['doc']
-)
+
+extras_require['dev'].extend(extras_require['test'])
+extras_require['dev'].extend(extras_require['lint'])
+extras_require['dev'].extend(extras_require['doc'])
+
```
cburgdorf: Mmmh...this is how we do it in all projects. Are you saying we should change this across the board in all repos? I don't have a strong opinion since it's only in the `setup.py` file but I want to bring up that:

1. This is more repetitive and imho harder to parse
2. If we adopt it, I think it should be adopted everywhere including our template project (paging in @carver for another opinion)

carver: We have been experimenting with other options, but definitely not settled on anything. Here is another approach that I think worked okay: https://github.com/ethereum/eth-abi/pull/130#discussion_r279082896

carver: Otherwise, I think the linter will be happy with just moving the `+` to the beginning of the line, which maybe seems cleaner than all the extends.

Or yet another option:
```py
extras_require['dev'].extend([
    extras_require['test'],
    extras_require['lint'],
    extras_require['doc'],
])

Comment (pipermerriam) [reply setup.py]: My linter doesn't like the `+` at the beginning or end... :frowning_face:

Comment (pipermerriam) [reply setup.py]: I have things I dislike about all of our approaches.... I'm fine with us normalizing this across all repos but I think we need to have a method we like before we can do that...

**Comment by pipermerriam** [comment]
Heavy handed merging.....

File: setup.py (hunk 1 above)
Comment (carver) [reply setup.py]: > My linter doesn't like the `+` at the beginning or end...

Which linter? It's pretty strange to not allow multiline addition...

Comment (pipermerriam) [reply setup.py]: synstastic in vim.  `W503` if it's before.  `W504` if it's after.

Comment (carver) [reply setup.py]: Yeah, seems like we should just do what flake8 does everywhere, until we decide to make portfolio-wide changes about how `flake8` works. IIRC, flake8 complains about trailing but not leading `+`s.

Comment (pipermerriam) [reply setup.py]: You are probably ignoring `W503` then if yours is not complaining about leading.

Comment (carver) [reply setup.py]: I don't have any local overrides. I had it backwards though:

```py
a = (
    1 +
    2 +
    3
)
```

Doesn't raise any `flake8` errors, using the `flake8` settings from py-evm, for example.

PR #60, opened 2019-05-14
Namespace legacy asyncio endpoint code and tests

pipermerriam:
## What was wrong?

The `asyncio` endpoint needs to make room for the incoming `trio` endpoint.

## How was it fixed?

Namespaced the `asyncio`  based `Endpoint` implementation to live under `lahja.asyncio` and the tests under `tests/core/asyncio`.

The name of the class is now `AsyncioEndpoint`.

There is a subclass that now lives at `lahja.endpoint.Endpoint` and `lahja.Endpoint` which issues a deprecation warning.

#### Cute Animal Picture

dscn0019_orig


lahja/asyncio/endpoint.py
```diff
@@ -528,3 +527,12 @@ def subscribe(self,
...
                     break
+
+
+class Endpoint(AsyncioEndpoint):
+    def __init__(self) -> None:
+        warnings.warn(
+            "`lahja.Endpoint` has been deprecated in favor of `lahja.AsyncioEndpoint`",
+            category=DeprecationWarning,
+        )
+        super().__init__()
```
cburgdorf: At this stage of the library, with us most likely being the only users, I think we don't need to issue deprecation warnings and just break things as we see the need for it.

tests/core/asyncio/test_basics.py
```diff
@@ -9,14 +9,17 @@
...
-    Endpoint,
     UnexpectedResponse,
 )
 
 
 @pytest.mark.asyncio
-async def test_request(endpoint):
+async def test_request(endpoint, event_loop):
+    import logging
+    logging.info('LOOP IN TEST: %s', id(event_loop))
```
cburgdorf: Was this meant to stay here? If yes, why not import `logging` at the top?

tests/core/asyncio/test_connect.py
```diff
@@ -4,17 +4,16 @@
...
-    Endpoint,
 )
 
 
 @pytest.mark.asyncio
 async def test_can_not_connect_conflicting_names():
-
     own = ConnectionConfig.from_name(generate_unique_name())
-    endpoint = Endpoint()
+    endpoint = AsyncioEndpoint()
```
cburgdorf: Not for this PR but in general: If the `TrioEndpoint` is API compatible to the `AsyncioEndpoint`. I guess instead of duplicating the tests we should be able to reuse much of them and just run them once with a `TrioEndpoint` and once with a `AsyncioEndpoint` right?

tests/core/asyncio/test_basics.py
pipermerriam: no, should be removed.

tests/core/asyncio/test_connect.py
pipermerriam: @cburgdorf non-trivial to do with the way that the `pytest-asyncio` and `pytest-trio` mark tests functions, but in general, that would be ideal.  Assuming we plan to drop the `asyncio` implementations then I think it's fine to maintain them independent of each other.  I'm really not sure what the right thing to do is at this stage.  Thoughts?

cburgdorf: Ah, alright. If there's no simple solution, I'm ok with copying the tests.

>Assuming we plan to drop the asyncio implementations then I think it's fine to maintain them independent of each other. I'm really not sure what the right thing to do is at this stage. Thoughts?

I feel we should not drop the `AsyncioEndpoint` and here is why: I believe (and I may be biased here) that lahja provides one of the best solutions to do efficient and ergonomic multiprocess communication in the current Python landscape. Trinity aside, I believe it is a net win for the project to support both endpoints. Especially if we consider that `asyncio` is the *standard* solution for asynchronous IO in Python (even though `trio` seems to be the *better*  one).

Also, supporting both kind of endpoints is pretty nice for migrating a code base from `asyncio` to `trio`. Today, it's only us using this library but this may change in the future.

I think we can revisit this if maintaining the `AsyncioEndpoint` ever becomes too much of a burden but for now I'd prefer that we keep it.

pipermerriam: :+1: we can decide this later once we're done with it and see if maintaining it adds too much overhead.

PR #66, opened 2019-05-20
[WIP] legacy asyncio endpoint reverse connection

pipermerriam:
This is a dirty WIP with a bunch of random stuff that I'll split out into many smaller PRs.  Speed optimizations, API optimizations and refactors, some bugfixes, more thorough testing of API features.

### DIFF
--- a/lahja/common.py
+++ b/lahja/common.py
@@ -1,26 +1,25 @@
 from abc import ABC, abstractmethod
 from pathlib import Path
-from typing import (  # noqa: F401
+from typing import (
+    TYPE_CHECKING,
     Any,
     Callable,
     Generic,
     NamedTuple,
     Optional,
+    Set,
     Type,
     TypeVar,
     Union,
 )
 
-
-class Subscription:
-    def __init__(self, unsubscribe_fn: Callable[[], Any]) -> None:
-        self._unsubscribe_fn = unsubscribe_fn
-
-    def unsubscribe(self) -> None:
-        self._unsubscribe_fn()
+if TYPE_CHECKING:
+    from lahja.base import EndpointAPI
 
 
 class BroadcastConfig:
+    __slots__ = ("filter_endpoint", "filter_event_id", "internal")
+
     def __init__(
         self,
         filter_endpoint: Optional[str] = None,
@@ -49,22 +48,34 @@ class BroadcastConfig:
 
 
 class BaseEvent:
+    __slots__ = ("_origin", "_id")
 
-    _origin = ""
-    _id: Optional[str] = None
-    _config: Optional[BroadcastConfig] = None
+    @property
+    def is_bound(self) -> bool:
+        return hasattr(self, "_origin")
 
-    def broadcast_config(self, internal: bool = False) -> BroadcastConfig:
+    def bind(self, endpoint: "EndpointAPI", id: Optional[str] = None) -> None:
+        """
+        Bind an event to an end
+        """
+        self._origin = endpoint.name
+        self._id = id
+
+    def response_config(self, internal: bool = False) -> BroadcastConfig:
         if internal:
             return BroadcastConfig(internal=True, filter_event_id=self._id)
-
-        return BroadcastConfig(filter_endpoint=self._origin, filter_event_id=self._id)
+        else:
+            return BroadcastConfig(
+                filter_endpoint=self._origin, filter_event_id=self._id
+            )
 
 
 TResponse = TypeVar("TResponse", bound=BaseEvent)
 
 
 class BaseRequestResponseEvent(ABC, BaseEvent, Generic[TResponse]):
+    __slots__ = ()
+
     @staticmethod
     @abstractmethod
     def expected_response_type() -> Type[TResponse]:
@@ -73,20 +84,47 @@ class BaseRequestResponseEvent(ABC, BaseEvent, Generic[TResponse]):
         This ensures that at runtime, only expected responses can be send
         back to callsites that issued a `BaseRequestResponseEvent`
         """
-        raise NotImplementedError("Must be implemented by subsclasses")
+        pass
 
 
-class TransparentEvent(BaseEvent):
+class SubscriptionsUpdated(NamedTuple):
+    subscriptions: Set[Type[BaseEvent]]
+    response_expected: bool
+
+
+class SubscriptionsAck:
+    __slots__ = ()
+
+
+class Message(ABC):
     """
-    This event is used to create artificial activity so that code that
-    blocks on a :meth:`~multiprocessing.queues.Queue.get` unblocks and
-    gets a chance to revalidate if it should continue to block for reading.
+    Base class for all valid message types that an ``Endpoint`` can handle.
+    ``NamedTuple`` breaks multiple inheritance which means, instead of regular subclassing,
+    derived message types need to derive from ``NamedTuple`` directly and call
+    Message.register(DerivedType) in order to allow isinstance(obj, Message) checks.
     """
 
     pass
 
 
-TRANSPARENT_EVENT = TransparentEvent()
+class Broadcast(NamedTuple):
+    event: Union[BaseEvent, bytes]
+    config: Optional[BroadcastConfig]
+
+
+Message.register(Broadcast)
+Message.register(SubscriptionsUpdated)
+Message.register(SubscriptionsAck)
+
+
+class Subscription:
+    __slots__ = ("_unsubscribe_fn",)
+
+    def __init__(self, unsubscribe_fn: Callable[[], Any]) -> None:
+        self._unsubscribe_fn = unsubscribe_fn
+
+    def unsubscribe(self) -> None:
+        self._unsubscribe_fn()
 
 
 class ConnectionConfig(NamedTuple):
@@ -107,8 +145,3 @@ class ConnectionConfig(NamedTuple):
             return cls(name=name, path=base_path / f"{name}.ipc")
         else:
             raise TypeError("Provided `base_path` must be a directory")
-
-
-class Broadcast(NamedTuple):
-    event: Union[BaseEvent, bytes]
-    config: Optional[BroadcastConfig]
--- a/lahja/common.py
+++ b/lahja/common.py
@@ -87,6 +87,21 @@ class BaseRequestResponseEvent(ABC, BaseEvent, Generic[TResponse]):
         pass
 
 
+class _MyNameIs(BaseEvent):
+    __slots__ = ("name",)
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+
+class _WhoAreYou(BaseRequestResponseEvent[_MyNameIs]):
+    __slots__ = ()
+
+    @staticmethod
+    def expected_response_type() -> Type[_MyNameIs]:
+        return _MyNameIs
+
+
 class SubscriptionsUpdated(NamedTuple):
     subscriptions: Set[Type[BaseEvent]]
     response_expected: bool
--- a/tests/core/asyncio/test_asyncio_endpoint_establishes_reverse_connection.py
+++ b/tests/core/asyncio/test_asyncio_endpoint_establishes_reverse_connection.py
@@ -1,0 +1,26 @@
+import asyncio
+
+import pytest
+
+from conftest import generate_unique_name
+from lahja import ConnectionConfig
+from lahja.endpoint import Endpoint
+
+
+@pytest.mark.asyncio
+async def test_legacy_server_endpoint_establishes_reverse_connection_to_client(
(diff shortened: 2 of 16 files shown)


lahja/common.py
```diff
@@ -0,0 +1,111 @@
...
+class BaseEvent:
+    __slots__ = ("_origin", "_id")
+
+    @property
+    def is_bound(self) -> bool:
+        return hasattr(self, '_origin')
+
+    def bind(self,
+             endpoint: 'EndpointAPI',
+             id: Optional[str] = None) -> None:
```
pipermerriam: @cburgdorf curious for feedback on this API.  I wanted to normalize the lifecycle of events, specifically setting of `_id` and `_origin`.  You can see some of this being used in the mechanisms for establishing reverse connections in the `_WhoAreYou` and `_MyNameIs` events.

lahja/common.py
```diff
@@ -0,0 +1,111 @@
...
+        """
+        Bind an event to an end
+        """
+        self._origin = endpoint.name
+        self._id = id
+
+    def response_config(self, internal: bool = False) -> BroadcastConfig:
```
pipermerriam: Wanting to rename this to `get_response_config`

lahja/common.py
```diff
@@ -0,0 +1,111 @@
...
+
+TResponse = TypeVar('TResponse', bound=BaseEvent)
+
+
+class BaseRequestResponseEvent(ABC, BaseEvent, Generic[TResponse]):
+    __slots__ = ()
+
+    @staticmethod
+    @abstractmethod
+    def expected_response_type() -> Type[TResponse]:
```
pipermerriam: @cburgdorf remind me why this was made a `staticmethod` rather than just an attribute/property

lahja/endpoint.py
```diff
@@ -118,13 +100,11 @@ def __init__(self, reader: StreamReader, writer: StreamWriter) -> None:
...
 
     async def send_message(self, message: Msg) -> None:
-        assert isinstance(message, Message)
         pickled = pickle.dumps(message)
         size = len(pickled)
 
         try:
-            self.writer.write(size.to_bytes(SIZE_MARKER_LENGTH, 'little'))
-            self.writer.write(pickled)
+            self.writer.write(size.to_bytes(SIZE_MARKER_LENGTH, 'little') + pickled)
```
pipermerriam: want to separate this out and profile it.

lahja/endpoint.py
```diff
@@ -140,43 +120,73 @@ def __init__(self, reader: StreamReader, writer: StreamWriter) -> None:
...
-            obj = pickle.loads(message)
-            assert isinstance(obj, Message)
+            obj = cast(Message, pickle.loads(message))
             return obj
         except (asyncio.IncompleteReadError, BrokenPipeError, ConnectionResetError):
             raise RemoteDisconnected()
 
 
-class InboundConnection:
+class RemoteEndpoint:
```
pipermerriam: @cburgdorf thoughts on this change, reverting back to a single connection type.  It seems necessary under the reverse connection paradigm since that allows for two way communication.

lahja/endpoint.py
```diff
@@ -190,61 +200,23 @@ def __init__(self, conn: Connection, new_msg_func: Callable[[Broadcast], None])
...
-        if not passes_config:
-            return False
+        if config is not None:
+            if config.filter_event_id is not None:
+                return True
+            elif not config.allowed_to_receive(self.name):
+                return False
 
-        return is_response or passes_filter
+        return type(item) in self.subscribed_messages
```
pipermerriam: Want to profile this as well, intended to optimize this function for early returns and not pre-computing un-needed information in those cases.

lahja/endpoint.py
```diff
@@ -266,35 +238,49 @@ class Endpoint(BaseEndpoint):
...
-    _internal_queue: 'asyncio.Queue[Tuple[BaseEvent, Optional[BroadcastConfig]]]'
-    _internal_loop_running: asyncio.Event
-
     _server_running: asyncio.Event
 
-    _loop: Optional[asyncio.AbstractEventLoop]
+    _loop: Optional[asyncio.AbstractEventLoop] = None
+
+    def __init__(self, name: str) -> None:
+        self.name = name
```
pipermerriam: @cburgdorf thoughts on this being core API? Core API is that endpoints **always** have a name.

lahja/endpoint.py
```diff
@@ -370,18 +342,25 @@ def name(self) -> str:  # type: ignore  # mypy thinks the signature does not mat
...
         self._receiving_queue.put_nowait((message.event, message.config))
 
     async def _accept_conn(self, reader: StreamReader, writer: StreamWriter) -> None:
         conn = Connection(reader, writer)
-        remote = InboundConnection(conn, self.receive_message)
-        self._inbound_connections.add(remote)
+        # We don't yet know the name of this remote so we use `None`.  If a
+        # successful reverse connection is established then the endpoint will
+        # be upgraded to a full connection and given a name.
+        remote = RemoteEndpoint(None, conn, self.receive_message)
```
pipermerriam: The `None` here is awkward.... still puzzling over an alternative.

lahja/endpoint.py
```diff
@@ -397,24 +376,29 @@ def subscribed_events(self) -> Set[Type[BaseEvent]]:
         Tell all inbound connections of our new subscriptions
         """
         # make a copy so that the set doesn't change while we iterate over it
-        for inbound_connection in self._inbound_connections.copy():
-            await inbound_connection.notify_subscriptions_updated(self.subscribed_events)
+        subscriptions = self.subscribed_events
+
+        await asyncio.gather(*(
+            remote.notify_subscriptions_updated(subscriptions)
```
pipermerriam: Noting that this doesn't expose the ability to make this operation block which might be desired.

lahja/endpoint.py
```diff
@@ -467,21 +451,13 @@ def _throw_if_already_connected(self, *endpoints: ConnectionConfig) -> None:
...
         while self._running:
             (item, config) = await self._receiving_queue.get()
+            self.logger.info('%s: PROCESSING ITEM FROM RECEIVING QUEUE: %s', self.name, item)
             try:
                 event = self._decompress_event(item)
                 self._process_item(event, config)
             except Exception:
                 traceback.print_exc()
 
-    async def _connect_internal_queue(self) -> None:
```
pipermerriam: @cburgdorf thoughts on this change.  Seems like a nice simplification to just process internal events immediately.

lahja/endpoint.py
```diff
@@ -266,35 +238,49 @@ class Endpoint(BaseEndpoint):
...
 
         self._child_tasks: Set['asyncio.Future[Any]'] = set()
 
         self._running = False
-        self._loop = None
-
-    @property
-    def ipc_path(self) -> pathlib.Path:
-        return self._ipc_path
+        self._serving = False
```
pipermerriam: This is a simple stand-alone, ensuring that running and serving are independent concepts.

lahja/endpoint.py
```diff
@@ -550,6 +527,68 @@ def _process_item(self, item: BaseEvent, config: Optional[BroadcastConfig]) -> N
...
+        self._who_are_you_stream_handler_running.set()
+        async for request in self.stream(_WhoAreYou):
+            await self.broadcast(_MyNameIs(self.name), request.response_config())
+
+    async def _establish_reverse_connection(self, remote: RemoteEndpoint) -> None:
+        item = _WhoAreYou()
+        # We bind the event as if it was a request/response, giving it an _id
+        # which allows the remote to respond to us without needing to receive a
+        # subscription update.
+        item.bind(self, str(uuid.uuid4()))
```
pipermerriam: TODO: profile and optimize if we can gain some speed by using a different source of unique ids.

lahja/endpoint.py
```diff
@@ -641,8 +705,7 @@ def broadcast_nowait(self,
         should be broadcasted to. By default, requests are broadcasted across all connected
         endpoints with their consuming call sites.
         """
-        item._origin = self.name
-        item._id = str(uuid.uuid4())
+        item.bind(self, str(uuid.uuid4()))
```
pipermerriam: TODO, I think the `id` should potentially be passed into `broadcast` to remove the *weirdness* of this pre-binding the event, and thus `broadcast` having to check if the event is already bound.

lahja/endpoint.py
```diff
@@ -729,4 +792,6 @@ def remove() -> None:
                     break
         finally:
             self._queues[event_type].remove(casted_queue)
+            if not self._queues[event_type]:
```
pipermerriam: This is a bug that should be backported, though I think I have a better solution using a combination of a `WeakValueDictionary` and a `WeakSet` that makes this auto-cleanup.

lahja/misc.py
```diff
@@ -18,6 +14,7 @@
 
 
 class Subscription:
+    __slots__ = ('_unsubscribe_fn',)
```
pipermerriam: TODO, extract all these `__slots__` and profile to see if they have a performance effect.

lahja/misc.py
```diff
@@ -53,53 +51,6 @@ def allowed_to_receive(self, endpoint: str) -> bool:
...
-    def expected_response_type() -> Type[TResponse]:
-        """
-        Return the type that is expected to be send back for this request.
-        This ensures that at runtime, only expected responses can be send
-        back to callsites that issued a `BaseRequestResponseEvent`
-        """
-        raise NotImplementedError("Must be implemented by subsclasses")
-
-
-class TransparentEvent(BaseEvent):
```
pipermerriam: Probably stand-alone if this removal doesn't break things.

tests/core/asyncio/test_asyncio_endpoint_establishes_reverse_connection.py
```diff
@@ -0,0 +1,26 @@
...
+from lahja import (
+    ConnectionConfig,
+)
+from lahja.endpoint import (
+    Endpoint,
+)
+
+
+@pytest.mark.asyncio
+async def test_legacy_server_endpoint_establishes_reverse_connection_to_client(ipc_base_path):
```
pipermerriam: TODO, rename `legacy` to `asyncio`

tests/core/asyncio/test_broadcast.py
```diff
@@ -0,0 +1,77 @@
...
+    await endpoint.start_serving(config)
+    # We need to connect to our own Endpoint if we care about receiving
+    # the events we broadcast. Many tests use the same Endpoint for
+    # broadcasting and receiving which is a valid use case so we hook it up
+    with endpoint:
+        yield endpoint
+
+
+@pytest.fixture(scope='function')
+async def client(server, config):
```
pipermerriam: TODO, drop these in favor of the `endpoint_pair` fixture that automatically tests both directions.

lahja/common.py
cburgdorf: Good question. I don't see it being accessed statically anywhere. I don't see a reason (maybe there was in the past?) why that shouldn't just become an abstract property

lahja/common.py
cburgdorf: Fine by me in general but maybe `create_response_config` or `derive_response_config` would reflect better what it does? We aren't just accessing any internal thing here but instead create a `BroadcastConfig` based on this event to form a reply. No strong feelings though.

lahja/endpoint.py
cburgdorf: I believe a scheme such as `endpoint_name` + timestamp + continuous integer should do it. The timestamp would be to get some extra safety. E.g. if we ever do what @lithp once suggested that we would save events into an internal database so that we could replay them, then just going with `endpoint_name` + continuous integer would probably not be enough to guarantee uniqueness.

lahja/common.py
cburgdorf: Makes sense. It always looked a bit odd that we were setting private properties.

lahja/misc.py
cburgdorf: What scares me about these slots is that they could get out of sync with the actual properties so easily. Same with typos, just feels very fragile. I wonder if something like https://github.com/cjrh/autoslot would make this cleaner?

pipermerriam: I don't think there's much to be scared of.  If there are extra slots then we're just taking a very small performance hit.  If there are properties not accounted for then we will get a very loud error trying to set that property.

cburgdorf: >If there are properties not accounted for then we will get a very loud error trying to set that property

Oh, nice! I didn't know they were all or nothing (but just tried it out to confirm). That makes them less scary indeed!

pipermerriam:
Replaced by #88

PR #68, opened 2019-05-20
Rename the `misc` module to `common`

pipermerriam:
## What was wrong?

The term miscellaneous is overly generic and has been bugging me a bit.

## How was it fixed?

Renamed it to `lahja.common` to represent these are the *common* datastructures that are shared between implementations of the endpoint concept.

#### Cute Animal Picture

mini-horse jpg 492x0_q85_crop-smart


cburgdorf:
Ah, and CI seems unhappy.

pipermerriam:
You might give me a few minutes on all of these, I'm in the process of splitting them up and then I know many of them will need individual care since my changes didn't really divide themselves up very easily into their individual concepts (lots of muddled changes that need to be manually unraveled)

pipermerriam:
Reset to master and just did this properly, purely a move and rename now.

PR #70, opened 2019-05-20
Update `EndpointAPI.subscribe` to be async and to accept coroutines

### LINKED ISSUE #65: Allow coroutines in `EndpointAPI.subscribe`
## What was wrong?

The `EndpointAPI.subscribe` API does not allow use of coroutines as handlers.  This restricts the api from being used as a building block for server-style request/response handlers since these will need to use the `EndpointAPI.broadcast` API.

While use of something like `EndpointAPI.broadcast_nowait` could be suitable, it seems more correct to allow coroutines.

## How can it be fixed?

Use [`inspect.iscoroutine`](https://docs.python.org/3.5/library/inspect.html#inspect.iscoroutine) to determine how the handler is called.

pipermerriam:
fixes #65 

## What was wrong?

The `EndpointAPI.subscribe` API was defined as synchronous, but has an internal need to perform an asynchronous action to update the endpoint subscriptions.

## How was it fixed?

Updated the API to be async as well as expanding the API to allow the *handler* to be a coroutine.

#### Cute Animal Picture

ZippyCowboy05-01-2013

### DIFF
--- a/lahja/base.py
+++ b/lahja/base.py
@@ -5,6 +5,7 @@ from typing import (  # noqa: F401
     Any,
     AsyncContextManager,
     AsyncGenerator,
+    Awaitable,
     Callable,
     Dict,
     Iterable,
@@ -55,6 +56,8 @@ class EndpointAPI(ABC):
     The :class:`~lahja.endpoint.Endpoint` enables communication between different processes
     as well as within a single process via various event-driven APIs.
     """
+
+    __slots__ = ("name",)
 
     name: str
 
@@ -184,10 +187,10 @@ class EndpointAPI(ABC):
         ...
 
     @abstractmethod
-    def subscribe(
+    async def subscribe(
         self,
         event_type: Type[TSubscribeEvent],
-        handler: Callable[[TSubscribeEvent], None],
+        handler: Callable[[TSubscribeEvent], Union[Any, Awaitable[Any]]],
     ) -> Subscription:
         """
         Subscribe to receive updates for any event that matches the specified event type.
--- a/tests/core/asyncio/test_asyncio_subscriptions_api.py
+++ b/tests/core/asyncio/test_asyncio_subscriptions_api.py
@@ -1,0 +1,223 @@
+import asyncio
+
+import pytest
+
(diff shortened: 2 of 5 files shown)


tests/core/asyncio/test_asyncio_subscriptions_api.py
```diff
@@ -0,0 +1,223 @@
...
+
+    async with AsyncioEndpoint.serve(config_a) as server_a:
+        async with AsyncioEndpoint.serve(config_b) as server_b:
+            async with AsyncioEndpoint.serve(config_c) as server_c:
+                async with AsyncioEndpoint("client").run() as client:
+                    await client.connect_to_endpoint(config_a)
+                    await client.connect_to_endpoint(config_b)
+                    await client.connect_to_endpoint(config_c)
+
+                    yield client, server_a, server_b, server_c
```
cburgdorf: If this was not just a fixture, I'd be concerned about the rightwards drift and recommend to combine this into a single `async_with` statement.


```python
    async with AsyncioEndpoint.serve(
        config_a
    ), AsyncioEndpoint.serve(
        config_b
    ), AsyncioEndpoint.serve(
        config_c
    ), AsyncioEndpoint("client").run()) as server_a, server_b, server_c, client:
        yield client, server_a, server_b, server_c
```

But given this is just a fixture with a single `yield` statement I think this is fine :+1:

lahja/base.py
```diff
@@ -184,10 +187,10 @@ def broadcast_nowait(
         ...
 
     @abstractmethod
-    def subscribe(
+    async def subscribe(
```
cburgdorf: One thing I just noticed: We *might* need a `subscribe_nowait` API now because the `subscribe` API could previously used from synchronous methods. Well there's a good chance that we might be able to convert these places into `async` methods but some could be tricky.

pipermerriam: I also realized that I'm using the wrong `inspect` function and that asynchronous subscription handlers don't appear to be working in some cases.  I have a fix in one of my branches that I'll open up shortly.

PR #72, opened 2019-05-21
Allow usage of parent event types in wait_until_* tools

cburgdorf:
## What was wrong?

The `wait_until_*` family of APIs do not allow matching against a parent event type. However, allowing that can be more convenient e.g.

```python
bus.wait_until_any_connections_subscribed_to(BaseEvent)
```
## How was it fixed?

Using `issubclass` for the type checking.

#### Cute Animal Picture

### DIFF
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -246,8 +246,11 @@ class OutboundConnection:
             await self._received_subscription.wait()
 
     async def wait_until_subscribed_to(self, event: Type[BaseEvent]) -> None:
-        while event not in self.subscribed_messages:
+        while not self.is_subscribed_to(event):
             await self.wait_until_subscription_received()
+
+    def is_subscribed_to(self, event_type: Type[BaseEvent]) -> bool:
+        return any(issubclass(event, event_type) for event in self.subscribed_messages)
 
 
 TFunc = TypeVar("TFunc", bound=Callable[..., Any])
@@ -408,7 +411,7 @@ class AsyncioEndpoint(BaseEndpoint):
             raise Exception("there are no outbound connections!")
 
         for outbound in self._outbound_connections.values():
-            if event in outbound.subscribed_messages:
+            if outbound.is_subscribed_to(event):
                 return
 
         coros = [


cburgdorf:
@pipermerriam I know you wanted to change this API as a whole, but I thought this change was small enough to put it off again. It makes the tests in Trinity more ergonomic.

cburgdorf:
>This change would make it impossible to know for sure that someone was actually subscribed to your event since it would return True if they were subscribed to anything in the inheritance chain but not actually tell you if the specific event you are querying actually has a subscriber.

Mmh...ok, this is true for events that have further subsclasses. But you would still be able to exclusively match agianst a specific event that is at the very bottom of the inheritance chain.

Ok, so let's look at the actual use case: Let's imagine you have a class that streams 3 different event types. Usually streaming of these 5 events is setup in the same tick so when you write your tests you only care about that there is *any* subscriber at all.

You could write your tests like this:

```python

def test_foo(bus):

await bus.wait_until_all_connections_subscribed_to(EventOne)
```

But picking `EventOne` feels like an arbitrary choice. Why not `EventTwo`? .

Or you could write

```python

def test_foo(bus):

await bus.wait_until_all_connections_subscribed_to(EventOne, EventTwo, EventThree)
```

But hat feels cumbersome to maintain. What do you suggest?

pipermerriam:
So I'm going to divide this into two problems.

This first, using `issubclass` which results in an API that has the property of giving false positives is something I'm :-1: on.

The second is an efficient way to determine if some/all of the connected endpoints is subscribed to one or more events.

I'm going to advocate that the current API of only exposing any/all is insufficient and that we should rework this to a single lower level API that all the other ones build from.

The following is from the top of my head and has plenty of room for improvement.  Merely meant to be demonstrative of the concept.

```python
def get_connected_endpoints(self) -> Tuple[str, ...]:
    # returns a tuple of the names of all connected endpoints
    ...

def get_remote_subscriptions(self) -> Tuple[Tuple[Optional[str], Set[Type[BaseEvent]]]:
    # returns a tuple of 2-tuples (<endpoint-name>, <subscriptions-set>).  
    # connections which don't have a name use `None`
    ...
```

This exposes the base primitives for doing any check you might want, providing access to the set of connected endpoint names and the full data set of connection subscription information.

These would go on the `EndpointAPI`.  Then we can move the implementations of the `wait_until_all/any` onto `BaseEndpoint` and for cases like what you're trying to do, you can experiment with the right API locally using the base primitive APIs and then contribute it back upstream once you have something you think is right.

Also, `get_connected_endpoints` API would allow the implementation of `is_connected_to` to be moved to `BaseEndpoint`.

cburgdorf:
I like that direction. 

>Then we can move the implementations of the `wait_until_all/any` onto `BaseEndpoint`

I guess, to make that possible each endpoint would also need to expose an event for when the subscriptions are updated. Otherwise, the only other option for the `BaseEndpoint` to implement the `wait_until_all/any` methods that I see would be polling which doesn't seem ideal.

But maybe we can just raise an internal event for that that the `BaseEndpoint` could `stream` in the `wait_until_all/any` APIs.

I'll try something out.

pipermerriam:
Yeah, use of internal events (not `asyncio.Event/trio.Event` for signaling seems like it could be a nice way to implement these APIs in a generic way.

PR #56, opened 2019-05-10
[WIP] Trio based implementation of the Endpoint API

pipermerriam:
blocked by #88 

## What was wrong?

WIP

Ugly things:

- `pytest-asyncio` doesn't play nice in the test suite so I ended up having to segregate them.

## How was it fixed?

Summary of approach.

#### Cute Animal Picture

put a cute animal picture link inside the parentheses

PR #74, opened 2019-05-21
More flexible wait APIs

cburgdorf:
## What was wrong?

The current APIs that let one wait on remote subscriptions aren't flexible enough and aren't generalized enough to be shared across the `TrioEndpoint`

See previous discussion: https://github.com/ethereum/lahja/pull/72#issuecomment-494425741

## How was it fixed?

1. Every specific endpoint has to implement the following API

```python
    @abstractmethod
    def get_connected_endpoints_and_subscriptions(
        self
    ) -> Tuple[Tuple[str, Set[Type[BaseEvent]]], ...]:
        """
        Return all connected endpoints and their event type subscriptions to this endpoint.
        """
        pass
```

@pipermerriam This is where this PR diverges from your proposal because I *think* this single API should be enough. If a connected remote endpoint does not hold event subscriptions it will still be listed here with an empty set. So, I don't think we need to have a separate API. And even if, this seperate API could be build inside the `BaseEndpoint` on top of this API.

2. Every endpoint needs to raise the internal `RemoteSubscriptionChanged` event if the known set of remote subscription changes. 

3. The `BaseEndpoint` defines four new APIs that are hence shared across all specific endpoints.

```python
    def is_remote_subscribed_to(
        self, remote_endpoint: str, event_type: Type[BaseEvent]
    ) -> bool:
        """
        Return ``True`` if the specified remote endpoint is subscribed to the specified event type
        from this endpoint. Otherwise return ``False``.
        """
  
    def is_any_remote_subscribed_to(self, event_type: Type[BaseEvent]) -> bool:
        """
        Return ``True`` if at least one of the connected remote endpoints is subscribed to the
        specified event type from this endpoint. Otherwise return ``False``.
  
    def are_all_remotes_subscribed_to(self, event_type: Type[BaseEvent]) -> bool:
        """
        Return ``True`` if every connected remote endpoint is subscribed to the specified event
        type from this endpoint. Otherwise return ``False``.
        """
  
    async def wait_until_remote_subscribed_to(
        self, remote_endpoint: str, event: Type[BaseEvent]
    ) -> None:
        """
        Block until the specified remote endpoint has subscribed to the specified event type
        from this endpoint.
        """

    async def wait_until_any_remote_subscribed_to(self, event: Type[BaseEvent]) -> None:
        """
        Block until any other remote endpoint has subscribed to the specified event type
        from this endpoint.
        """

    async def wait_until_all_remotes_subscribed_to(
        self, event: Type[BaseEvent]
    ) -> None:
        """
        Block until all currently connected remote endpoints are subscribed to the specified
        event type from this endpoint.
        """
```

#### Cute Animal Picture

### DIFF
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -48,6 +48,7 @@ from lahja.common import (
     ConnectionConfig,
     Message,
     Msg,
+    RemoteSubscriptionChanged,
     Subscription,
     SubscriptionsAck,
     SubscriptionsUpdated,
@@ -246,10 +247,6 @@ class RemoteEndpoint:
         async with self._received_subscription:
             await self._received_subscription.wait()
 
-    async def wait_until_subscribed_to(self, event: Type[BaseEvent]) -> None:
-        while event not in self.subscribed_messages:
-            await self.wait_until_subscription_received()
-
 
 @asynccontextmanager  # type: ignore
 async def run_remote_endpoint(remote: RemoteEndpoint) -> AsyncIterable[RemoteEndpoint]:
@@ -440,41 +437,16 @@ class AsyncioEndpoint(BaseEndpoint):
         for remote in tuple(self._full_connections.values()):
             await remote.notify_subscriptions_updated(subscribed_events)
 
-    async def wait_until_any_connection_subscribed_to(
-        self, event: Type[BaseEvent]
-    ) -> None:
+    def get_connected_endpoints_and_subscriptions(
+        self
+    ) -> Tuple[Tuple[Optional[str], Set[Type[BaseEvent]]], ...]:
         """
-        Block until any other endpoint has subscribed to the ``event`` from this endpoint.
+        Return all connected endpoints and their event type subscriptions to this endpoint.
         """
-        if len(self._full_connections) == 0:
-            raise Exception("there are no outbound connections!")
-
-        for outbound in self._full_connections.values():
-            if event in outbound.subscribed_messages:
-                return
-
-        coros = [
-            outbound.wait_until_subscribed_to(event)
+        return tuple(
+            (outbound.name, outbound.subscribed_messages)
             for outbound in self._full_connections.values()
-        ]
-        _, pending = await asyncio.wait(coros, return_when=asyncio.FIRST_COMPLETED)
-        (task.cancel() for task in pending)
-
-    async def wait_until_all_connections_subscribed_to(
-        self, event: Type[BaseEvent]
-    ) -> None:
-        """
-        Block until all other endpoints that we are connected to are subscribed to the ``event``
-        from this endpoint.
-        """
-        if len(self._full_connections) == 0:
-            raise Exception("there are no outbound connections!")
-
-        coros = [
-            outbound.wait_until_subscribed_to(event)
-            for outbound in self._full_connections.values()
-        ]
-        await asyncio.wait(coros, return_when=asyncio.ALL_COMPLETED)
+        )
 
     def _compress_event(self, event: BaseEvent) -> Union[BaseEvent, bytes]:
         if self.has_snappy_support:
@@ -563,7 +535,15 @@ class AsyncioEndpoint(BaseEndpoint):
         self._full_connections[config.name] = remote
 
         task = asyncio.ensure_future(self._handle_server(remote))
+        subscriptions_task = asyncio.ensure_future(
+            self.watch_outbound_subscriptions(remote)
+        )
+        subscriptions_task.add_done_callback(self._endpoint_tasks.remove)
+        self._endpoint_tasks.add(subscriptions_task)
+
         task.add_done_callback(self._endpoint_tasks.remove)
+        task.add_done_callback(lambda _: subscriptions_task.cancel())
+
         self._endpoint_tasks.add(task)
 
         # don't return control until the caller can safely call broadcast()
@@ -576,6 +556,13 @@ class AsyncioEndpoint(BaseEndpoint):
         finally:
             if remote.name is not None:
                 self._full_connections.pop(remote.name)
+
+    async def watch_outbound_subscriptions(self, outbound: RemoteEndpoint) -> None:
+        while outbound in self._full_connections.values():
+            await outbound.wait_until_subscription_received()
+            await self.broadcast(
+                RemoteSubscriptionChanged(), BroadcastConfig(internal=True)
+            )
 
     def is_connected_to(self, endpoint_name: str) -> bool:
         return endpoint_name in self._full_connections
--- a/lahja/base.py
+++ b/lahja/base.py
@@ -27,6 +27,7 @@ from .common import (
     ConnectionConfig,
     Message,
     Msg,
+    RemoteSubscriptionChanged,
     Subscription,
 )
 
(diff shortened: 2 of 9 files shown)


lahja/base.py
```diff
@@ -200,3 +193,86 @@ def __reduce__(self) -> None:  # type: ignore
...
+        """
+        Return ``True`` if the specified remote endpoint is subscribed to the specified event type
+        from this endpoint. Otherwise return ``False``.
+        """
+        for endpoint, subscriptions in self.get_connected_endpoints_and_subscriptions():
+            if endpoint != remote_endpoint:
+                continue
+
+            for subscribed_event_type in subscriptions:
+                if subscribed_event_type == event_type:
```
pipermerriam: `is` checks for direct class comparison are slightly more correct since classes are singletons.

lahja/asyncio/endpoint.py
```diff
@@ -534,16 +506,30 @@ def connect_to_endpoints_nowait(self, *endpoints: ConnectionConfig) -> None:
...
         self._child_tasks.add(task)
 
         # don't return control until the caller can safely call broadcast()
         await remote.wait_until_subscription_received()
 
+    async def watch_outbound_subscriptions(self, outbound: OutboundConnection) -> None:
+        while outbound in self._outbound_connections.values():
+            await outbound.wait_until_subscription_received()
+            await self.broadcast(
+                RemoteSubscriptionChanged(), BroadcastConfig(internal=True)
```
pipermerriam: I'm questioning this API though I'm not at a point where I'm sure.  I was thinking that rather than specifying this mechanism, we could instead have two more core API: `EndpointAPI.wait_until_remote_subscriptions_change()` and `EndpointAPI.wait_until_local_subscriptions_change()` and it's up to the endpoint to decide on implementation details.

One reason for leaning this direction is that I *think* some of these might be more appropriately implemented using `asyncio.Condition` since we have multiple things waiting on the same resource (subscriptions changing).

cc @lithp does this look like the right use case for the `Condition` API?

cburgdorf: >One reason for leaning this direction is that I think some of these might be more appropriately implemented using `asyncio.Condition` 

The underlying `wait_until_subscription_received` is already implemented using `asyncio.Condition` so I think everything should be sound.

I think the reason I like this more is that instead of just having an API that can *block* you up until the next change, this gives you an API that you can continuously get notified for updates with.

I think that makes it more flexible to build up on.

tests/core/asyncio/test_subscription_wait_apis.py
```diff
@@ -0,0 +1,69 @@
+import asyncio
+
+import pytest
+
+from helpers import DummyRequestPair, DummyResponse
+
+
+@pytest.mark.asyncio
+async def test_wait_until_any_all_subscribed(triplet_of_endpoints):
```
cburgdorf: @pipermerriam did you see these tests? I think they cover all the new APIs.

pipermerriam: I did, though I didn't see that they hit all the APIs.  However, I'm inclined to request a test per unit of API functionality.  Under this structure we avoid a future situation where a refactor accidentally removes coverage of an API.  With very granular tests that each try to test a single thing, it becomes much easier to understand what each test is intended to test.

lahja/base.py
```diff
@@ -200,3 +193,86 @@ def __reduce__(self) -> None:  # type: ignore
...
+        self, event: Type[BaseEvent]
+    ) -> None:
+        """
+        Block until all currently connected remote endpoints are subscribed to the specified
+        event type from this endpoint.
+        """
+
+        if self.are_all_remotes_subscribed_to(event):
+            return
+
```
lithp: I think there's a race-condition if we receive `RemoteSubscriptionChanged` right here. It'll be dropped and then this method will block until the next `RemoteSubscriptionChanged` comes in, which might take a while.

cburgdorf: But how would that happen? We aren't using multiple threads (are we?) and a context switch can only happen at the next `await`, no?

lithp: You're right. I looked more carefully and there's *almost* a race-condition because `stream` blocks on notifying before it starts reading from the queue, but it adds the queue first so you're right that there are no switch points between the check above this and waiting for the message.

lahja/asyncio/endpoint.py
lithp: I'd argue that neither is more flexible; given one you can easily build the other:

```python
async def stream():
  while True:
    await wait_until_subscription_received()
    yield RemoteSubscriptionChanged()

async def wait_until_subscription_received():
  async for _ in stream():
    return
```

This PR makes it look like the `stream()` variant is useful more often so probably that's the better interface to expose!

At first glance this *does* look like a good use-case for `Condition`. This design is certainly better than `wait_until_any_connection_subscribed_to` was and I can't figure out why it feels weird to me, but I think there's a chance that using a `Condition` would simplify this code even more. I think it'd involve the `Endpoint` holding onto a condition and passing it into each of the `OutboundConnection`s, they'd notify it where they're already notifying `_received_subscriptions` then Endpoint's `wait_until_any_outbound_receives_subscription` could be used to succinctly build the rest of these methods.

lahja/base.py
pipermerriam: I've been tinkering in another branch and have a deadlock that I'm trying to unravel.  Makes me want a solid test suite around these APIs because it seem to be quite easy to get into a situation where we can deadlock like this.

lahja/asyncio/endpoint.py
cburgdorf: >given one you can easily build the other

You totally can, but I would still say it is easier to build on top of the event API. E.g you don't even need to implement the `wait_until_subscription_received` because you can do this already.

```
await bus.wait_for(RemoteSubscriptionChanged)
```

I would say, that's the beauty of the dogfooding approach: It works with `stream`, `wait_for` and `subscribe` out of the box.

lahja/base.py
cburgdorf: >That's assuming that we only switch when the co-routine actually blocks, which I think is the default asyncio behavior

Correct. I just refreshed my memory ([see this](https://stackoverflow.com/questions/54860285/how-can-i-prevent-context-switching-when-calling-an-async-function/54863862#54863862) and [that](https://djangostars.com/blog/asynchronous-programming-in-python-asyncio/)) and this is how it is supposed to work.

>In the unlikely event that uvloop could switch away at any `await` call then I think there's still a race-condition, because there's an implicit await when we loop over `self.stream`

Correct, but as far as I understand things (and it really makes my head spin!), this should not be the case.

@pipermerriam 

>I've been tinkering in another branch and have a deadlock that I'm trying to unravel

Do you have a link for the curious?

At the risk of pointing out the obvious, you do use `PYTHONASYNCIODEBUG=1` for debugging these, right? I've found that to be a real life safer.

pipermerriam: It's trio stuff and it's in #56 and nowhere near ready for anything beyond curious observers.  Still major structural changes happening as I re-shape the core `asyncio` library implementation to a form that's close enough to the future `trio` implementation that they can properly interop without gross hacks.

tests/core/asyncio/test_subscription_wait_apis.py
cburgdorf: Moved them into separate tests :+1:

cburgdorf:
@pipermerriam I rebased this and addressed the feedback. Do we want to take it or would you like to explore other possible APIs?

lahja/base.py
```diff
@@ -248,3 +241,86 @@ def __reduce__(self) -> None:  # type: ignore
...
+        """
+        for endpoint, subscriptions in self.get_connected_endpoints_and_subscriptions():
+            if endpoint != remote_endpoint:
+                continue
+
+            for subscribed_event_type in subscriptions:
+                if subscribed_event_type is event_type:
+                    return True
+
+        return False
```
pipermerriam: Wondering if we should error in this case (that no endpoint name matched the provided name).  Seems like an easy mistake to make.

cburgdorf: I was wondering the same but here is why I didn't do that:

1. If we look at the `wait_until_*` family of APIs, I think it is totally reasonable to start waiting on remote subscriptions even *before* a specific connection has been established. I'm trying to see this from a broader non-trinity based perspective but even *within* Trinity I can see that being useful. E.g. Hot-plugin replacement/loading could cause new endpoints to come up later in the lifecycle of the application and having this raise an exception may make that harder.

2. Of course, we could say we change `is_remote_subscribed_to` to raise an exception and then have the `wait_until_*` APIs catch and ignore this so that only the `is_remote_subscribed_to` API raises but I think that would introduce an inconsistency between these APIs with little gain in return.

PR #76, opened 2019-05-21
Fix warnings during shutdown

pipermerriam:
## What was wrong?

Lots of warnings during test suite (and really any Endpoint shutdown)

## How was it fixed?

- Catch the `RuntimeError` caused by the event loop being closed.
- Manually manage the background tasks.

#### Cute Animal Picture

shutterstock_144263860-e1497461501753


cburgdorf:
I'm going to merge this because I *think* it might remove some flakiness from the `test_exceptions_dont_stop_processing` test.

PR #77, opened 2019-05-21
Error check in main connect_to_endpoint API

pipermerriam:
## What was wrong?

The `Endpoint.connect_to_endpoint` API was not checking for an existing connection.

## How was it fixed?

Added the check and updated test to exercise that code path.

#### Cute Animal Picture

Muffin-The-Mini-Mule-A-Cr-006


pipermerriam:
@lithp if you don't object a lot of these cleanups are actually pre-requisites to a much larger refactor to the entire test suite to better cover all of the formal APIs more clearly and explicitely so I'm inclined to accept the drop in coverage in favor expanding it very soon in a much more deliberate way (approximately late next week).

PR #78, opened 2019-05-21
BaseEvent.bind API

pipermerriam:
## What was wrong?

Both the `broadcast` and `request` methods were setting private attributes of the `BaseEvent`

## How was it fixed?

Added `BaseEvent.bind` API to deal with these and internal `_broadcast` implementation that does the binding in a common location.

#### Cute Animal Picture

RedPom2012

### DIFF
--- a/lahja/common.py
+++ b/lahja/common.py
@@ -12,6 +12,8 @@ from typing import (  # noqa: F401
     Union,
 )
 
+from lahja.exceptions import BindError
+
 if TYPE_CHECKING:
     from lahja.base import EndpointAPI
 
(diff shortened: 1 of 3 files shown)


lahja/common.py
```diff
@@ -52,7 +56,15 @@ class BaseEvent:
 
     _origin = ""
     _id: Optional[str] = None
-    _config: Optional[BroadcastConfig] = None
+
+    is_bound = False
+
+    def bind(self, endpoint: "EndpointAPI", id: Optional[str]) -> None:
+        if self.is_bound:
+            raise RuntimeError("Event is already bound")
```
cburgdorf: Not a strong opinion but maybe worth following our standard practice to use a subclass of `LahjaError` here.

PR #79, opened 2019-05-21
Add broadcast_nowait to core API

### LINKED ISSUE #64: Make `broadcast_nowait` a core API
### What is wrong

`Endpoint.broadcast_nowait()` is likely a necessary API.  After spending some time with it I'm convinced that `broadcast_nowait` should be a core event API.  

### How can it be fixed.

Add it to the core `EndpointAPI` definition.

Supporting it in trio should be fine, it will just involve having an extra background process that receive events that have been placed on the a `SendChannel` using the [`SendChannel.send_nowait`](https://trio.readthedocs.io/en/latest/reference-io.html#trio.abc.SendChannel.send_nowait) API.

pipermerriam:
fixes #64

## What was wrong?

It turns out that `Endpoint.broadcast_nowait` is likely necessary.  Without it you simply cannot talk to the event bus from APIs that are synchronous.

## How was it fixed?

No core change other than adding it to the core API definition.

> I plan to majorly expand the test suite soon but only after I get the reverse server -> client connection stuff in because it improves the testing experience significantly

#### Cute Animal Picture

Cute-Goats-Photo-06

### DIFF
--- a/lahja/base.py
+++ b/lahja/base.py
@@ -140,6 +140,25 @@ class EndpointAPI(ABC):
         pass
 
     @abstractmethod
+    def broadcast_nowait(
+        self, item: BaseEvent, config: Optional[BroadcastConfig] = None
+    ) -> None:
+        """
+        A sync compatible version of :meth:`~lahja.base.EndpointAPI.broadcast`
+
+        .. warning::
+
+            Heavy use of :meth:`~lahja.base.EndpointAPI.broadcast_nowait` in
+            contiguous blocks of code without yielding to the `async`
+            implementation should be expected to cause problems.  Without
+            yielding, the braodcasted events will typically build back-pressure
+            either in the form of unbounded memory growth or errors due to full
+            queues.
+
+        """
+        pass
+
+    @abstractmethod
     async def request(
         self,
         item: BaseRequestResponseEvent[TResponse],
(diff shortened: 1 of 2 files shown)


lahja/base.py
```diff
@@ -121,6 +121,25 @@ def is_connected_to(self, endpoint_name: str) -> bool:
...
+    ) -> None:
+        """
+        A sync compatible version of :meth:`~lahja.base.EndpointAPI.broadcast`
+
+        .. warning::
+
+            Heavy use of :meth:`~lahja.base.EndpointAPI.broadcast_nowait` in
+            contiguous blocks of code without yielding to the `async`
+            implementation should be expected to cause problems.  Without
+            yielding, the braodcasted events will typically build back-pressure
```
lithp: nitpick: the problem is that they'll fail to provide back-pressure! Maybe: "the broadcasted events will fail to provide backpressure and queue up until control is returned back to the event loop. This could result in unbounded memory growth or errors due to full queues"

cburgdorf: There's also a typo there: `braodcasted` -> `broadcasted`

pipermerriam: I decided to just remove the last sentence since each implementation could have it's own problems dependning on how it's implemented.

PR #82, opened 2019-05-21
optimize can_send_item a little

pipermerriam:
## What was wrong?

The `OutboundConnection` object's `can_send_item` method checked *up-to* three things to determine if sending was valid.  The current implementation computed all three of these conditions and then iterated them in precedence order to determine the result.

This is sub-optimal because in cases where the function returns because of the highest precedence check, the lower precedence checks are not used.

## How was it fixed?

Adjust the function such that it only computes the minimal data it needs.

#### Cute Animal Picture

miniature-hereford


pipermerriam:
Profile showed about a 1% gain so it wasn't a lot but it wasn't trivial... adding comment

PR #85, opened 2019-05-24
Combine inbound and outbound connection concepts

pipermerriam:
built on top of #70 

## What was wrong?

The concept op remote connections is currently divide across two classes.  One for *inbound* and one for *outbound*.  The reality is these connections are bi-directional, we just don't take advantage of this.

It is currently not intuitive that if I do `client.connect_to_endpoint(server)` and then `server.broadcast()` that the client will not receive the broadcast.

For connections that need to talk both ways, leveraging both directions of the connection should reduce the number of sockets and connections we need to make.

## How was it fixed?

In the upcoming `trio` implementation in #56 there is an API that leverages the bi-directional nature of connections to make `server.broadcast()` send things back to clients that are connected.

This PR is a prerequisite to that functionality, collapsing the connection concept down to a single class.  This PR makes it *capable* of using the connection in both directions but does not actually do anything with the other side of the connection yet.

Summary of approach.

#### Cute Animal Picture

5f1db895145dab887f00602e4401dd6f--baby-horses-mini-horses

### DIFF
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -76,7 +76,7 @@ SIZE_MARKER_LENGTH = 4
 
 
 class Connection(ConnectionAPI):
-    logger = logging.getLogger("lahja.endpoint.Connection")
+    logger = logging.getLogger("lahja.endpoint.asyncio.Connection")
 
     def __init__(self, reader: StreamReader, writer: StreamWriter) -> None:
         self.writer = writer
@@ -131,7 +131,7 @@ class RemoteEndpoint:
             an event meant to be processed by the endpoint.
     """
 
-    logger = logging.getLogger("lahja.endpoint.InboundConnection")
+    logger = logging.getLogger("lahja.endpoint.asyncio.RemoteEndpoint")
 
     def __init__(
         self,
@@ -262,8 +262,9 @@ SubscriptionSyncHandler = Callable[[BaseEvent], Any]
 
 class AsyncioEndpoint(BaseEndpoint):
     """
-    The :class:`~lahja.asyncio.AsyncioEndpoint` enables communication between different processes
-    as well as within a single process via various event-driven APIs.
+    The :class:`~lahja.endpoint.asyncio.AsyncioEndpoint` enables communication
+    between different processes as well as within a single process via various
+    event-driven APIs.
     """
 
     _ipc_path: Path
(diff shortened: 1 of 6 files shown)


lahja/asyncio/endpoint.py
```diff
@@ -141,10 +187,16 @@ def __init__(
                 return
 
             if isinstance(message, Broadcast):
-                self.new_msg_func(message)
+                await self.new_msg_func(message)
```
pipermerriam: This change is effctively changing us from using `Queue.put_nowait` to `Queue.put` which seems more appropriate (but I don't really know why).  Thoughts?

lahja/asyncio/endpoint.py
```diff
@@ -111,25 +115,70 @@ def __init__(self, reader: StreamReader, writer: StreamWriter) -> None:
...
+    - ``SubscriptionsUpdated``
+            broadcasting the subscriptions that the endpoint on the other side
+            of this connection is interested in.
+    - ``SubscriptionsAck``
+            acknowledgedment of a ``SubscriptionsUpdated``
+    - ``Broadcast``
+            an event meant to be processed by the endpoint.
     """
 
+    logger = logging.getLogger("lahja.endpoint.InboundConnection")
```
cburgdorf: `"lahja.endpoint.InboundConnection"` -> `"lahja.endpoint.RemoteEndpoint`"

lahja/asyncio/endpoint.py
```diff
@@ -111,25 +115,70 @@ def __init__(self, reader: StreamReader, writer: StreamWriter) -> None:
...
+
     def __init__(
-        self, conn: Connection, new_msg_func: Callable[[Broadcast], None]
+        self,
+        name: Optional[str],
```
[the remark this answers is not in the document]
pipermerriam: The `name` is an implicit part of the connection.  When a server receives an incoming connection over the socket it doesn't know the name of the endpoint on the other side of the connection.  This is the concept of a *half* connection.

I however now realize, that with subscriptions, we can potentially get rid of the distinction between half/full connections since we still have a mechanism for determining which endpoints are acceptable to send over.

However, I am going to open an issue to deal with this separately.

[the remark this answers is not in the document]
pipermerriam: https://github.com/ethereum/lahja/issues/96

lithp:
Thanks for the ping but no complaints here! I had originally broken them into two classes because I thought that made it easier to reason about how endpoints communicate with inbound and outbound peers, but I agree that it was wasteful to use 2x the number of connections and if merging them makes trio easier I'm all in :+1:

lahja/asyncio/endpoint.py
```diff
@@ -591,13 +665,13 @@ def stop(self) -> None:
...
-            self._process_item(item, config)
+            await self._process_item(item, config)
             return
 
         # Broadcast to every connected Endpoint that is allowed to receive the event
         compressed_item = self._compress_event(item)
         disconnected_endpoints = []
-        for name, remote in list(self._outbound_connections.items()):
+        for name, remote in list(self._full_connections.items()):
             # list() makes a copy, so changes to _outbount_connections don't cause errors
```
lithp: This comment should be updated to say `_full_connections`. My careless typo probably foiled your search-replace :sweat_smile:

PR #86, opened 2019-05-24
Add `EndpointAPI.wait_until_connected_to` 

pipermerriam:
- builds on #85 

## What was wrong?

Needed a non-polling mechanism to wait for a connection.

## How was it fixed?

Additive changes to our connections are now guarded with an `asyncio.Condition`.  This allows implementation of `wait_until_connected_to` in a non-polling manner.  After the initial connection check, the API waits for an update to the connection before checking again.

#### Cute Animal Picture

b97196bbcd1219819dbbd3aaf74c8847


pipermerriam:
I've got some PR stacking going on here.  This contains the diff from 3 PRs, only latest commit is relevant.

lithp:
Don't know whether you intentionally didn't use this, but if you push the first two commits to this repo under a branch you can then set the [base branch](https://help.github.com/en/articles/changing-the-base-branch-of-a-pull-request) of this PR so that the diff only shows the last commit. Just have to remember to set the base back to master once the preceding PRs have been merged!

pipermerriam:
@lithp since this stretches across forks, that results in the PR being only open in my fork which lacks visibility.  Maybe I should just open these as drafts so that there's less confusion and they can be reviewed sequentially as they get merged and rebased.

tests/core/asyncio/test_asyncio_wait_until_connected_to.py
```diff
@@ -0,0 +1,19 @@
...
+
+import pytest
+
+from lahja import AsyncioEndpoint, ConnectionConfig
+
+
+@pytest.mark.asyncio
+async def test_connect_to_endpoint(ipc_base_path):
+    config = ConnectionConfig.from_name("server", base_path=ipc_base_path)
+    async with AsyncioEndpoint.serve(config) as server:
```
lithp: Love this API :heart_eyes:

lahja/asyncio/endpoint.py
```diff
@@ -403,12 +422,18 @@ def run(self, *args, **kwargs):  # type: ignore
...
             async with run_remote_endpoint(remote):
-                await remote.wait_stopped()
+                try:
+                    await remote.wait_stopped()
+                except asyncio.CancelledError:
+                    return
         finally:
-            self._half_connections.remove(remote)
+            if remote in self._half_connections:
+                self._half_connections.remove(remote)
```
lithp: It might be a little cleaner to replace this finally block with another `task.add_done_callback(self._half_connections.remote(remote))` inside `_accept_conn`.

lahja/asyncio/endpoint.py
```diff
@@ -403,12 +422,18 @@ def run(self, *args, **kwargs):  # type: ignore
...
+        await self._add_half_connection(remote)
+
     async def _handle_client(self, remote: RemoteEndpoint) -> None:
         try:
             async with run_remote_endpoint(remote):
-                await remote.wait_stopped()
+                try:
+                    await remote.wait_stopped()
+                except asyncio.CancelledError:
+                    return
```
lithp: Why does this suppress `CancelledError`? I think that in this situation it's okay but letting it propagate up is safer.

From [the docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.Task):
> cancelled() can be used to check if the Task was cancelled. The method returns True if the wrapped coroutine did not suppress the CancelledError exception and was actually cancelled.

Suppressing the error prevents callers of `_handle_client` which didn't use `ensure_future` from knowing they've been cancelled, and it also prevents other callers of `_handle_client` from determining whether the coro stopped because of cancellation. Neither kind of caller exists right now but this still seems like unexpected behavior.

pipermerriam:
@lithp I had to make some structural changes to fix a race condition where subscription updates could end up not transmitting if they occured at a specific timing during the connection.  Would like one more pass if you will.

lahja/asyncio/endpoint.py
```diff
@@ -153,6 +153,12 @@ def __init__(
         self._running = asyncio.Event()
         self._stopped = asyncio.Event()
 
+    def __str__(self) -> str:
+        return f"RemoteEndpoint[{self.name if self.name is not None else id(self)}]"
+
+    def __repr__(self) -> str:
+        return f"<{self}>"
```
pipermerriam: can/should extract to standalone.

lahja/asyncio/endpoint.py
```diff
@@ -192,11 +198,11 @@ def __init__(
                 async with self._received_response:
                     self._received_response.notify_all()
             elif isinstance(message, SubscriptionsUpdated):
-                self.subscribed_messages = message.subscriptions
                 async with self._received_subscription:
+                    self.subscribed_messages = message.subscriptions
                     self._received_subscription.notify_all()
-                if message.response_expected:
-                    await self.send_message(SubscriptionsAck())
+                    if message.response_expected:
```
pipermerriam: @lithp can you confirm this should be how this is structured or do you think it's ok to send the `Ack` after we've released the lock...

lahja/asyncio/endpoint.py
```diff
@@ -235,13 +245,12 @@ def can_send_item(self, item: BaseEvent, config: Optional[BroadcastConfig]) -> b
...
 
-    async def wait_until_subscription_received(self) -> None:
-        async with self._received_subscription:
-            await self._received_subscription.wait()
-
     async def wait_until_subscribed_to(self, event: Type[BaseEvent]) -> None:
-        while event not in self.subscribed_messages:
-            await self.wait_until_subscription_received()
+        async with self._received_subscription:
+            while True:
```
pipermerriam: @lithp can you confirm this is ok according to your understanding of conditions.  rather than release and then re-acquire at the entrance to each loop, we eliminate this and only switch on the release/acquire that happens as part of the `wait` call.

lahja/asyncio/endpoint.py
```diff
@@ -360,16 +382,77 @@ def run(self, *args, **kwargs):  # type: ignore
...
+        await asyncio.gather(
+            *(
+                remote.notify_subscriptions_updated(subscribed_events)
+                for remote in itertools.chain(
+                    self._half_connections.copy(), self._full_connections.values()
+                )
+            )
+        )
+
+    async def _connect_receiving_queue(self) -> None:
```
pipermerriam: Didn't intend to relocate this but I think I like it here better.  Need to add some comments to deliniate the body of this class more clearly but I'll do that as a cleanup PR since re-organizing makes code hard to audit.

TODO: revert this move.

lahja/asyncio/endpoint.py
```diff
@@ -360,16 +382,77 @@ def run(self, *args, **kwargs):  # type: ignore
...
+        self._subscription_updates_running.set()
+        async with self._subscription_updates_condition:
+            while self.is_running:
+                await self._subscription_updates_condition.wait()
+                await self._do_notify_subscriptions_changed()
+
+    def _notify_subscriptions_changed_nowait(self) -> None:
+        self._subscription_updates_queue.put_nowait(None)
+
+    async def _do_notify_subscriptions_changed(self) -> None:
```
pipermerriam: TODO: remove this method and inline it to `_process_subscription_updates` since we never want anyone to call this directly.

lahja/asyncio/endpoint.py
```diff
@@ -360,16 +382,77 @@ def run(self, *args, **kwargs):  # type: ignore
...
+    async def _process_subscription_updates_queue(self) -> None:
+        while self.is_running:
+            await self._subscription_updates_queue.get()
+            async with self._subscription_updates_condition:
+                self._subscription_updates_condition.notify_all()
+
+    async def _process_subscription_updates(self) -> None:
+        self._subscription_updates_running.set()
+        async with self._subscription_updates_condition:
+            while self.is_running:
```
pipermerriam: TODO: invert the `while` and lock acquisition to reduce lock release/acquisition churn.

lahja/asyncio/endpoint.py
```diff
@@ -360,16 +382,77 @@ def run(self, *args, **kwargs):  # type: ignore
...
 
+    async def _process_subscription_updates_queue(self) -> None:
+        while self.is_running:
+            await self._subscription_updates_queue.get()
+            async with self._subscription_updates_condition:
+                self._subscription_updates_condition.notify_all()
+
+    async def _process_subscription_updates(self) -> None:
+        self._subscription_updates_running.set()
+        async with self._subscription_updates_condition:
```
pipermerriam: @lithp Needed this to ensure we don't end up with concurrent subscription updates being sent out since it's possible that they arrive out of order in that case and then the other side will have potentially in-accurate subscription records.

TOOD: add a code comment explaining this.

lahja/tools/benchmark/process.py
```diff
@@ -42,21 +44,34 @@ def start(self) -> None:
 
     def stop(self) -> None:
         assert self._process is not None
-        self._process.terminate()
-        self._process.join(1)
+        if self._process.pid is not None:
+            os.kill(self._process.pid, signal.SIGINT)
+        else:
+            self._process.terminate()
```
pipermerriam: TODO: extract to standalone.

lahja/asyncio/endpoint.py
lithp: I think it's fine to send the `Ack` from outside the lock's block, and maybe even preferable! The Condition is there to coordinate the local process, since the remote process doesn't even know about this Condition I don't think that releasing the lock first could create any race conditions.

lahja/asyncio/endpoint.py
lithp: This looks good to me!

lahja/asyncio/endpoint.py
```diff
@@ -216,9 +222,13 @@ def __init__(
         endpoint has acknowledged the new subscription set. If ``block`` is ``False`` then this
         function will return immediately after the send finishes.
         """
```
lithp: The comment just above this line isn't quite right anymore: "Alert the ``Endpoint`` which has connected to us that our subscription set has changed." Since `RemoteEndpoint` now handles both inbound and outbound this should be something like "Alert the remote that this endpoint's subscription set has changed"

lahja/asyncio/endpoint.py
```diff
@@ -368,16 +390,77 @@ def run(self, *args, **kwargs):  # type: ignore
     async def start(self) -> None:
         if self.is_running:
             raise RuntimeError(f"Endpoint {self.name} is already running")
+
         self._receiving_loop_running = asyncio.Event()
         self._receiving_queue = asyncio.Queue()
 
+        self._subscription_updates_running = asyncio.Event()
+        self._subscription_updates_condition = asyncio.Condition()
+        self._subscription_updates_queue = asyncio.Queue()
```
lithp: These could be initialized in `__init__` which would save you from needing to declare them at the beginning of this class and would also make it easier to reason about the potential for `None` dereferences.

The `await *.running()` pattern sounds like it could be pulled out into a helper which removes the potentially `None` instance attribute entirely. Something like `await self.start_coro(self._process_subscription_updates)` which doesn't return until the coro has started.

lahja/asyncio/endpoint.py
```diff
@@ -242,13 +252,12 @@ def can_send_item(self, item: BaseEvent, config: Optional[BroadcastConfig]) -> b
...
-        while event not in self.subscribed_messages:
-            await self.wait_until_subscription_received()
+        async with self._received_subscription:
+            while True:
+                if event in self.subscribed_messages:
+                    return
+                await self._received_subscription.wait()
 
 
 @asynccontextmanager  # type: ignore
```
lithp: As of this change `_handle_client` and `_handle_server` are the same method and together they constitute the only callers of `run_remote_endpoint`. Maybe they could all be merged into one method (untested):

```python
async def run_remote_endpoint(remote: RemoteEndpoint) -> None
    await remote.start()
    try:
        await remote.wait_stopped()
    finally:
        await remote.stop()
```

lahja/asyncio/endpoint.py
lithp: How would they arrive out of order? Domain sockets always deliver messages in order and `_notify_lock` ensures there's only one update per socket happening at any time.

lahja/asyncio/endpoint.py
```diff
@@ -402,21 +485,29 @@ def run(self, *args, **kwargs):  # type: ignore
...
+        # the time that we manually update this individual connection and that
+        # we place it within the set of tracked connections, otherwise, a
+        # subscription update from elsewhere can occur between the time these
+        # two statements execute resulting in the remote missing a new
+        # subscription update.  Note that inverting these statements should
+        # also mitigate this, but it has the downside of the manual update
+        # potentially being redundant.
+        async with self._subscription_updates_condition:
+            await remote.notify_subscriptions_updated(self.subscribed_events)
+            await self._add_half_connection(remote)
```
lithp: Ahh, okay, I see the problem now, `_add_half_connection` blocks when it tries to acquire `connections_changed` so there's a period in time where the remote isn't receiving subscription updates. Double-sending the same set of subscriptions has a very small cost though! And I think that trying not to send it introduced greater inefficiencies elsewhere, serializing all updates isn't cheap!

pipermerriam:
I've been futzing with this code for days so it's a bit muddled in my head but one thing that makes me prefer this approach is that it I'd "correct" in the sense that it properly maintains the necessary resource locks. The alternative one of double sending is likely also adequate in fixing the bug but it does so through some insurrection which I think means more mental overhead since it's no longer trivial to do things like grep the code for all of the places the resource is modified by looking for the places it gets locked.

Sent from ProtonMail mobile

-------- Original Message --------

lahja/asyncio/endpoint.py
pipermerriam: good catch.

lahja/asyncio/endpoint.py
pipermerriam: I'm going to leave this for now and open an issue to track as I've seen this issue arising as well but I don't want to conflate this PR with an extra fix that I'm not sure I know the right solution for.

pipermerriam: https://github.com/ethereum/lahja/issues/106

lahja/asyncio/endpoint.py
pipermerriam: (I thought I responded to this somewhere but I don't see it)

I'm inclined to leave it this way because it *correctly* manages the appropriate locks on the resource, where-as the other mechanism solves the problem through indirection.  First, I think the overall cost here is low because our usage patterns for the library don't involve adding new connections at a high frequency.  Second, doing it the alternate way of inverting these statements seems like a solution that could easily be undone incidentally in a refactor since it's an implicit solution.

lahja/asyncio/endpoint.py
pipermerriam: I started undoing this but then stopped as I'm inclined to leave it this way.  Here's my reasoning.

1. we end up needing to do subscription updates from synchronous methods with the addition of `subscribe_nowait` which seems to be a necessary API. 
2. I'm much more comfortable with that method doing a `Queue.put_nowait` than I am adding an additional call to `asyncio.ensure_future(self._notify_subscription_updates())`.
3. I think that the `asyncio.Condition` based approach is more appropriate and under that approach we still need a background process to listen for changes.  The various branches that I have in progress have made good use of that `Condition` API.

I'm fine unrolling this if you're still not convinced at the point where the `trio` based endpoint is being added to the library.

pipermerriam:
@lithp lets do a call later today when we're both online and see if we can find something that works for both of us.

- We need something that can work from synchronous methods as both `Subscription.remove` and `EndpointAPI.subscribe_nowait` need to be able to trigger subscription updates.
- I don't like using `asyncio.ensure_future(...)` as the mechanism for handling the synchronous context problem.
- I am pretty sure that a `Condition` based approach is justified as I've spent some serious time debugging race conditions over the last week around subscriptions not being properly updated and while I know we *can* achieve working code without the locks, I'm not excited about such solutions.

pipermerriam:
replaced by #110 and #108

PR #88, opened 2019-05-24
Make `AsyncioEndpoint` establish reverse connections

pipermerriam:
Builds on #86 

## What was wrong?

Currently, endpoints behave like this.

```python
server = ...
client = ...
client.connect_to_endpoint(server)
client.broadcast(...)  # the server will receive this.
server.broadcast(...)  # the client will NOT receive this
```

This is surprising behavior.

## How was it fixed?

Connections are now always bi-directional.  The `RemoteEndpoint` concept now enforces an internal `Hello` message to be delivered by each side of the connection to identify the name it should use with the connection.

This has some very nice properties that shake out:

- remote endpoints are now always named.
- we handle the `remote` the same way on both sides of the connection.

Other changes:

- All remotes now share a single `Condition` for signaling that subscriptions have changed in the remote endpoints, fixing a race condition where endpoints don't *notice* that one of their remotes is subscribed to something.
- Remotes now only expose a `wait_for_subscription_initialization` which replaces the imprecise use of `wait_until_subscription_received` which has the potential to miss the initial update if things happen *quickly*.
- It is no longer for an endpoint to connect to itself since it directly conflicts with the concept of establishing the reverse connection.

#### Cute Animal Picture

cute-baby-hippos-6-590841d1ae669__700


pipermerriam:
@lithp or @cburgdorf this is ready for preliminary review.  The benchmarks are timing out which I need to diagnose and fix but the core of the logic is not likely to change.

pipermerriam:
Pulled out: https://github.com/ethereum/lahja/pull/116 and #115 , probably able to extract a bit more, but have to step away for a bit.

lahja/asyncio/endpoint.py
```diff
@@ -851,7 +811,7 @@ def _remove_sync_subscription(
         A handler is passed as a second argument an :class:`~lahja.common.Subscription` is returned
         to unsubscribe from the event if needed.
         """
-        if inspect.iscoroutine(handler):
+        if inspect.iscoroutinefunction(handler):
```
lithp: [The docs](https://docs.python.org/3/library/asyncio-task.html#asyncio.iscoroutinefunction) say that `asyncio.iscoroutinefunction()` covers one extra case which I think is the legacy way coroutines used to be implemented? It sounds slightly better to use the asyncio version just in case users are still writing coroutines the old way.

lahja/asyncio/endpoint.py
```diff
@@ -439,22 +350,44 @@ def run(self, *args, **kwargs):  # type: ignore
...
-        # the Endpoint on the other end blocks until it receives this message
-        await remote.notify_subscriptions_updated(self.subscribed_events)
+        await remote.wait_started()
 
-    async def _handle_client(self, remote: RemoteEndpoint) -> None:
+        self._server_tasks.add(
+            asyncio.ensure_future(self._establish_reverse_connection(remote))
+        )
+
+    async def _establish_reverse_connection(self, remote: RemoteEndpoint) -> None:
```
lithp: This is only called from inside `_accept_conn` and could be in-lined into it.

lithp:
Looking at the `_WhoAreYou`, `_MyNameIs` messages there's some extra code around handling the request/response that isn't strictly necessary. In the same way that `SubscriptionsUpdated` is now sent immediately after a connection is established, `_MyNameIs` could be sent immediately after connecting to a remote. `_accept_conn` could block until that message is received and I think a few coros and a few dozen lines would be saved.

Alternatively, I'm pretty sure that with unix sockets you can call `bind()` before you call `connect()` and give your socket a meaningful name. I think this is [the `local_addr` parameter of create_connection](https://docs.python.org/3/library/asyncio-eventloop.html#asyncio.loop.create_connection), and the chosen name will show up as `peername` when [`get_extra_info()`](https://docs.python.org/3/library/asyncio-protocol.html#asyncio.BaseTransport.get_extra_info) is called on the writer. This would allow signaling the source endpoint without sending any messages at all! The higher-level trio APIs don't seem to support this but it provides the full set of low-level APIs, so it's possible to build bind-before-connect ourselves.

pipermerriam:
Lemme try your fancy socket magic

pipermerriam:
@lithp take a look at 14ebc77  it still uses an "internal" message type, but it avoids the round trip, cuts down on complexity, and has the nice benefit of `RemoteEndpoint.name` being strictly a `str` and no longer ever having a `None` value.

lahja/asyncio/endpoint.py
```diff
@@ -125,61 +125,28 @@ def __init__(self, reader: StreamReader, writer: StreamWriter) -> None:
...
-            of this connection is interested in.
-    - ``SubscriptionsAck``
-            acknowledgedment of a ``SubscriptionsUpdated``
-    - ``Broadcast``
-            an event meant to be processed by the endpoint.
-    """
-
-    logger = logging.getLogger("lahja.endpoint.asyncio.RemoteEndpoint")
-
+class RemoteEndpoint(BaseRemoteEndpoint):
```
lithp: The goal behind breaking this up might be more clear if it was called `AsyncioRemoteEndpoint`.

lithp:
Thanks for making that change, I think this is a nice improvement!

lahja/asyncio/endpoint.py
```diff
@@ -125,61 +125,28 @@ def __init__(self, reader: StreamReader, writer: StreamWriter) -> None:
...
     def __init__(
         self,
-        name: Optional[str],
         conn: Connection,
+        subscriptions_changed: ConditionAPI,
         new_msg_func: Callable[[Broadcast], Awaitable[Any]],
```
lithp: I'm not sure what should be done about the `asyncio.Condition` and `asyncio.Event` members, but `new_msg_func` can definitely go into the `BaseRemoteEndpoint` constructor!

lahja/asyncio/endpoint.py
```diff
@@ -653,11 +523,22 @@ def is_connected_to(self, endpoint_name: str) -> bool:
 
         if event_type in self._sync_handler:
```
lithp: Man, coroutines and mypy are pretty frustrating to work with. You have to write the exception handling logic twice for no good reason, and the naive way of getting around this is frustrated by mypy:

```python
async def call_handler(handler, item, is_sync: bool) -> None:
  try:
    if is_sync:
      handler(item)
    else:
      await handler(iteem)
  except Exception:
    self.logger.debug('Error in subscription handler")

if event_type in self._sync_handler:
    for handler in self._sync_handler[event_type]:
      await call_handler(handler, item, is_sync=True)

if event_type in self._async_handler:
  await asyncio.gather(
    call_handler(handler, item, is_sync=False)
    for handler in self._async_handler[event_type]
  )
```

lahja/asyncio/endpoint.py
```diff
@@ -718,13 +599,14 @@ def stop(self) -> None:
...
         """
-        await self._broadcast(item, config, None)
+        await self._broadcast(item, config, None, None)
 
     async def _broadcast(
         self,
         item: BaseEvent,
         config: Optional[BroadcastConfig],
         id: Optional[RequestID],
+        remotes_for_broadcast: Optional[Tuple[RemoteEndpointAPI, ...]],
```
lithp: Unless I'm missing something, this last parameter of `_broadcast` is only `not None` when the last parameter (`remotes_for_broadcast`) of `_request` is `not None`, but the only place where `_request` is called always passes in `None`. 

Is this leftover code, or maybe you plan on using it in a future PR?

lahja/asyncio/endpoint.py
```diff
@@ -735,22 +617,36 @@ def stop(self) -> None:
             return
 
         # Broadcast to every connected Endpoint that is allowed to receive the event
+        if remotes_for_broadcast is None:
```
lithp: nitpick: This might be cleaner as:

```python
if remotes_for_broadcast is None:
  remotes_for_broadcast = (
    remote for remote in self._connections
    if remote.can_send_item(item, config)
  )

await self._send_to_remotes(item, config, *remotes_for_broadcast)
```

lahja/asyncio/endpoint.py
```diff
@@ -735,22 +617,36 @@ def stop(self) -> None:
...
-                for name in disconnected_endpoints:
-                    self._full_connections.pop(name, None)
-                self._remote_connections_changed.notify_all()
+        message = Broadcast(compressed_item, config)
+        for remote in remotes:
+            try:
+                await remote.send_message(message)
+            except RemoteDisconnected:
+                self.logger.debug(f"Remote endpoint {remote} no longer connected")
+                remote.stop_nowait()
```
lithp: I think `stop_nowait()` could be dropped and this call turned into a call to `stop()` (which doesn't need to be async). Right now it's unclear which one you should call in different situations, and `stop_nowait()` knows a *lot* about the caller, it only works because of how `_run_remote_endpoint()` is written.

lahja/base.py
```diff
@@ -51,6 +57,228 @@ class ConnectionAPI(ABC):
...
+
+    async def wait_until_subscription_received(self) -> None:
+        ...
+
+
+class BaseRemoteEndpoint(RemoteEndpointAPI):
+    logger = logging.getLogger("lahja.base.RemoteEndpoint")
+
+    def __str__(self) -> str:
+        return f"RemoteEndpoint[{self.name if self.name is not None else id(self)}]"
```
lithp: This throws an attribute error now, I think it should read from `_name`.

lahja/asyncio/endpoint.py
pipermerriam: Yeah, I realized that a bunch of this machinery can now be removed since we no longer need to use these internals.

examples/inter_process_ping_pong.py
```diff
@@ -66,32 +59,29 @@ def run_proc2():
...
 
 
 async def display_proc1_events(endpoint):
     async for event in endpoint.stream(FirstThingHappened):
         logging.info("Received via STREAM API in proc2: %s", event.payload)
 
 
-# Helper function to send events every n seconds
-def is_nth_second(interval):
-    return int(time.time()) % interval is 0
```
pipermerriam: This didn't appear to be necessary since we can accomplish roughly the same thing using `await asyncio.sleep(interval)`

examples/request_api.py
```diff
@@ -47,14 +44,14 @@ def run_proc2():
 
 
 async def proc2_worker():
-    config = ConnectionConfig.from_name("e2")
-    async with AsyncioEndpoint.serve(config) as endpoint:
-        await endpoint.connect_to_endpoints(ConnectionConfig.from_name("e1"))
```
pipermerriam: The pattern of having an endpoint connect to itself is no longer valid under this new paradigm.  cc @cburgdorf

lahja/base.py
```diff
@@ -157,38 +158,122 @@ class BaseRemoteEndpoint(RemoteEndpointAPI):
...
+                # exited the code which actually updates the subscriptions and
+                # we are merely responding to the sender to acknowledge
+                # receipt.
+                if message.response_expected:
+                    await self.send_message(SubscriptionsAck())
+            elif isinstance(message, SubscriptionsAck):
+                async with self._received_response:
+                    self._received_response.notify_all()
+            else:
+                self.logger.error(f"received unexpected message: {message}")
```
pipermerriam: The body of this `while` loop is unchanged, only moved from it's previous location in `lahja.endpoint.asyncio`

pipermerriam:
@lithp I agree with you conceptually but this PR introduces what is effectively a backwards incompatible change of endpoints no longer being able to connect to themselves which changed how all of the tests operated since many of them used a single *self-connected* endpoint.

lithp:
@pipermerriam Are you sure that endpoints are no longer able to connect to themselves? I was able to get a test working where it seemed to work, and in fact where it needed to happen in order for the test to pass. Re the tests I think they're now missing a test that an Endpoint can connect to itself and send messages. This is important because the idea of loose-coupling means the caller shouldn't have to care where a given subscriber is; if the subscriber is on the same event bus as the caller the message should still be delivered! Currently, that requires connecting to yourself. I'd be all for dropping that requirement and having messages be propagated internally by default, but until then there should be a test for it.

lahja/base.py
```diff
@@ -157,38 +158,122 @@ class BaseRemoteEndpoint(RemoteEndpointAPI):
     bi-directional with messages flowing in either direction.
     """
 
-    logger = logging.getLogger("lahja.endpoint.asyncio.RemoteEndpoint")
+    logger = logging.getLogger("lahja.base.RemoteEndpoint")
```
lithp: While you're here maybe `AsyncioEndpoint` should also get a logger, right now I think that `self.logger` calls from inside it will use this one, which might be confusing.

lithp:
> there are maybe ~100 lines that I think I can get out of this diff but that still leaves it as +500/-400. I'd really like to move this forward but I'm also fine spending more time breaking it down if you really think its necessary.

It's okay, don't worry about breaking it up more, I've already partially reviewed most of this, but for future PRs I'll always be able to review a few smaller ones faster than a PR with them all combined!

tests/core/asyncio/test_broadcast_config.py
```diff
@@ -1,61 +1,70 @@
...
+    pass
 
-@pytest.mark.asyncio
-async def test_broadcasts_to_specific_endpoint(triplet_of_endpoints):
 
-    endpoint1, endpoint2, endpoint3 = triplet_of_endpoints
+@pytest.mark.asyncio
+async def test_broadcasts_to_specific_endpoint(server_with_two_clients):
 
-    tracker = Tracker()
```
lithp: `Tracker` is unused now, it can be removed with this PR.

tests/core/asyncio/test_broadcast_config.py
```diff
@@ -1,61 +1,70 @@
...
 
+class TailEvent(BaseEvent):
+    pass
 
-@pytest.mark.asyncio
-async def test_broadcasts_to_specific_endpoint(triplet_of_endpoints):
 
-    endpoint1, endpoint2, endpoint3 = triplet_of_endpoints
+@pytest.mark.asyncio
+async def test_broadcasts_to_specific_endpoint(server_with_two_clients):
```
lithp: I like this, it's an improvement over how the test was! However, there are now no tests for `Endpoint.request(..., BC(filter_endpoint=...))`, that sounds worth testing, maybe there could be an additional test after this one which tests the same thing but while making a request?

lahja/asyncio/endpoint.py
```diff
@@ -240,12 +200,17 @@ def __init__(self, name: str) -> None:
...
+        self._remote_connections_changed = asyncio.Condition()  # type: ignore
+
+        # Signal when at least one remote has had a subscription change.
+        self._remote_subscriptions_changed = asyncio.Condition()  # type: ignore
 
         # storage containers for inbound and outbound connections to other
         # endpoints
-        self._full_connections = {}
-        self._half_connections = set()
+        self._connections = set()
```
lithp: *nit*: `is_connected_to` and `_get_remote` would both be easier to write if this was a `Dict[remote.name, remote]`. I'm not sure whether that's worth making the *other* lines which use `_connections` more complicated but wanted to point it out.

lahja/asyncio/endpoint.py
```diff
@@ -494,68 +444,68 @@ def connect_to_endpoints_nowait(self, *endpoints: ConnectionConfig) -> None:
...
         finally:
-            if remote.name is not None:
+            await remote.stop()
+            if remote in self._connections:
                 async with self._remote_connections_changed:
-                    self._full_connections.pop(remote.name)
+                    self._connections.remove(remote)
                     self._remote_connections_changed.notify_all()
+                async with self._remote_subscriptions_changed:
+                    self._remote_subscriptions_changed.notify_all()
```
lithp: For symmetry with `self._add_connection` it might look better to pull this block out into a `self._remove_connection`.

lithp:
Although, one last thing, the current commit title isn't very informative, before merging it'd be nice for the commit to have something more descriptive of all the changes!

lahja/asyncio/endpoint.py
pipermerriam: Yup, that change works now that remotes always have names at the time we put them in this data structure which wasn't previously the case.  Will change.

pipermerriam:
Will get the final cleanups in place tomorrow, cleanup the commit, verify what the deal is with *self* connections and document it in a new issue if something needs to be done (I now agree that it makes sense to support it) and get this shipped, allowing us to cut `0.13.0` tomorrow if it's done early enough in the day.

examples/request_api.py
cburgdorf: But we are still able to subscribe to events that originate from our very own endpoint, right? Because I think that is pretty essential.

pipermerriam: Yes, I'm going to fix this.

lahja/asyncio/endpoint.py
pipermerriam: I'm actually not going to do this.  We can tackle it later if it proves to be a performance issue, however, I have grown to like that we have a single canonical location for endpoint names and to store it as a dict would allow for the `remote.name` and the name it's stored under to diverge.

PR #98, opened 2019-05-27
Stop using uuid for request_id

pipermerriam:
Builds on #99 

## What was wrong?

The cost of generating a UUID is *maybe* more expensive.

## How was it fixed?

Custom scheme that *might* have minimal overhead.

- [x] TODO: benchmark suite does not flex the `request/response` API.

#### Cute Animal Picture

35843c422aaa96acff54869fa97ef8e4

### DIFF
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -18,6 +18,7 @@ from typing import (  # noqa: F401
     Callable,
     DefaultDict,
     Dict,
+    Iterable,
     List,
     NamedTuple,
     Optional,
@@ -807,11 +808,13 @@ class AsyncioEndpoint(BaseEndpoint):
         self._queues[event_type].append(casted_queue)
         await self._notify_subscriptions_changed()
 
+        iterations: Iterable[int]
+
         if num_events is None:
             # loop forever
-            iterations = itertools.repeat(True)
+            iterations = itertools.count()
         else:
-            iterations = itertools.repeat(True, num_events)
+            iterations = range(num_events)
 
         try:
             for _ in iterations:
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -19,6 +19,7 @@ from typing import (  # noqa: F401
     DefaultDict,
     Dict,
     Iterable,
+    Iterator,
     List,
     NamedTuple,
     Optional,
@@ -29,7 +30,6 @@ from typing import (  # noqa: F401
     Union,
     cast,
 )
-import uuid
 
 from async_generator import asynccontextmanager
 
@@ -50,6 +50,7 @@ from lahja.common import (
     Message,
     Msg,
     RemoteSubscriptionChanged,
+    RequestIDGenerator,
     Subscription,
     SubscriptionsAck,
     SubscriptionsUpdated,
@@ -59,6 +60,7 @@ from lahja.exceptions import (
     RemoteDisconnected,
     UnexpectedResponse,
 )
+from lahja.typing import RequestID
 
 
 async def wait_for_path(path: Path, timeout: int = 2) -> None:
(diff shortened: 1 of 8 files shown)


pipermerriam:
Benchmarking suggest this is indeed measurably faster.  Cuts total time generating request id in the `request` method from being around `1.3%` to be `0.2%`.

lahja/asyncio/endpoint.py
```diff
@@ -277,7 +279,7 @@ class AsyncioEndpoint(BaseEndpoint):
     _receiving_queue: "asyncio.Queue[Tuple[Union[bytes, BaseEvent], Optional[BroadcastConfig]]]"
     _receiving_loop_running: asyncio.Event
 
-    _futures: Dict[Optional[str], "asyncio.Future[BaseEvent]"]
```
lithp: I'm surprised it was even possible for the key of a dictionary to be optional :smile:

lahja/asyncio/endpoint.py
```diff
@@ -286,17 +288,25 @@ class AsyncioEndpoint(BaseEndpoint):
     _sync_handler: DefaultDict[Type[BaseEvent], List[SubscriptionSyncHandler]]
 
     _loop: Optional[asyncio.AbstractEventLoop] = None
+    _get_request_id: Iterator[RequestID]
 
     def __init__(self, name: str) -> None:
         self.name = name
 
+        try:
+            self._get_request_id = RequestIDGenerator(name.encode("ascii") + b":")
```
lithp: nitpick: In an imaginary future where there are multiple backends users might want to switch between different kinds of endpoints. If this check is on the Asyncio endpoint but not the Trio one then non-ascii users could have a hard time experimenting with Asyncio after building their app with Trio.

Putting this check into `BaseEndpoint` would make life easier for those hypothetical users.

lahja/asyncio/endpoint.py
```diff
@@ -734,7 +747,9 @@ def broadcast_nowait(
 
         return result
 
-    def _remove_cancelled_future(self, id: str, future: "asyncio.Future[Any]") -> None:
+    def _remove_cancelled_future(
+        self, id: RequestID, future: "asyncio.Future[Any]"
+    ) -> None:
```
lithp: nitpick: Is this Black? I think our convention has been to not add this final newline

lahja/base.py
```diff
@@ -219,6 +219,12 @@ class BaseEndpoint(EndpointAPI):
 
     logger = logging.getLogger("lahja.endpoint.Endpoint")
 
+    def __str__(self) -> str:
+        return f"Endpoint[{self.name}]"
+
+    def __repr__(self) -> str:
+        return f"<{self.name}>"
```
lithp: nitpick: I think these two are flipped, [`__repr__`](https://docs.python.org/3/reference/datamodel.html#object.__repr__) is the one which roughly looks like a constructor. What's the rationale for implementing both? I think `__str__` is usually added if `__repr__` gives output which is too long, but here they both give short and printable output.

lahja/asyncio/endpoint.py
pipermerriam: Just mean `None` as a key which is possible, but you can only ever have one.

lahja/asyncio/endpoint.py
pipermerriam: :+1:

lahja/asyncio/endpoint.py
pipermerriam: black

lahja/asyncio/endpoint.py
cburgdorf: This is one of the few occasions where `Union[str, None]` (which is equivalent to `Optional[str]`) may *feel* a little more suitable :sweat_smile:

PR #101, opened 2019-05-27
Fix examples and cover them in CI

cburgdorf:
## What was wrong?

1. The examples in the repo did not work anymore
2. The examples where not covered during CI

## How was it fixed?

1. Fixed the examples
2. Added tests so that they don't break again
3. Added new `py36-examples` and `py37-examples` jobs to cover these tests.

#### Cute Animal Picture

### DIFF
--- a/examples/request_api.py
+++ b/examples/request_api.py
@@ -3,7 +3,7 @@ import multiprocessing
 import time
 
 from lahja import (
-    Endpoint,
+    AsyncioEndpoint,
     BaseEvent,
     BaseRequestResponseEvent,
     BroadcastConfig,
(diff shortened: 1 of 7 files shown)


cburgdorf:
@pipermerriam CI doesn't seem to kick in here? Not sure if it is related to the issue that was already reported yesterday.

cburgdorf:
Looks like CI works again. :tada:

examples/request_api.py
```diff
@@ -53,23 +47,24 @@ def run_proc2():
...
+        for i in range(3):
+            print("Requesting")
+            result = await endpoint.request(GetSomethingRequest())
+            print(f"Got answer: {result.payload}")
+
 
 if __name__ == "__main__":
 
-    multiprocessing.set_start_method('spawn')
+    multiprocessing.set_start_method("spawn")
```
pipermerriam: It might be worth commenting *why* on this assuming it's necessary for things to work properly.

PR #102, opened 2019-05-28
Does it trigger CI?

cburgdorf:
## What was wrong?

Issue #

## How was it fixed?

Summary of approach.

#### Cute Animal Picture

### DIFF
--- a/examples/inter_process_ping_pong.py
+++ b/examples/inter_process_ping_pong.py
@@ -1,52 +1,61 @@
 import asyncio
 import multiprocessing
 import time
+import sys
 
-from lahja import (
-    BaseEvent,
-    Endpoint,
-    ConnectionConfig,
-)
+from lahja import BaseEvent, AsyncioEndpoint, ConnectionConfig
+
+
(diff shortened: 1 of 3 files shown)


examples/inter_process_ping_pong.py
```diff
@@ -1,52 +1,61 @@
...
-)
+from lahja import BaseEvent, AsyncioEndpoint, ConnectionConfig
 
 
-class BaseExampleEvent(BaseEvent):
+# The example in this file runs infinitely until the user stops the program.
+# To be able to test the example and run assertions on the live output, we
+# need to ensure the output won't get buffered.
+def print_line(line, *args):
+    print(line, *args, flush=True)
```
pipermerriam: logging hooked up to stderr might be roughly as good as this with the benefit of promoting good design patterns.

cburgdorf: It didn't occur to me that using logging would fix that but I tried it out and you are right! Plus, I refactored the code to use a proper formatter with timestamps to get rid of the ugly ad hoc timestamps that were used before :)

PR #105, opened 2019-05-29
Add subscribe_nowait method

cburgdorf:
## What was wrong?

The `subscribe`  API is often used as an escape hatch to subscribe to events from synchronous methods. Recently that method became `async` by default and looking into Trinity there seem to be a lot of places that would probably need to broader refactoring to turn them into `async` methods.

## How was it fixed?

Since Trinity is currently on a lahja git ref, my main motivation here is to give us something that would quickly resolve the situation so that we can cut a new release. This adds a new API `subscribe_nowait` that can be used as a drop in replacement for the old API. Notice that it still has the additional feature of being able to accept async handlers.

Also this fixes the issue of not using `iscoroutinefunction` for the handler check.

I intentionally did not add this to the `EndpointAPI` as I'm not sure if we want to support it long term.

#### Cute Animal Picture

### DIFF
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -841,15 +841,13 @@ class AsyncioEndpoint(BaseEndpoint):
         # that's a performance problem, not a correctness problem.
         self._subscriptions_changed.set()
 
-    async def subscribe(
+    def subscribe_nowait(
         self,
         event_type: Type[TSubscribeEvent],
         handler: Callable[[TSubscribeEvent], Union[Any, Awaitable[Any]]],
     ) -> Subscription:
         """
-        Subscribe to receive updates for any event that matches the specified event type.
-        A handler is passed as a second argument an :class:`~lahja.common.Subscription` is returned
-        to unsubscribe from the event if needed.
+        A sync compatible version of :meth:`~lahja.asyncio.AsyncioEndpoint.subscribe`
         """
         if inspect.iscoroutine(handler):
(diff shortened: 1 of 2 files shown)


pipermerriam:
I think we do want to add this to `EndpointAPI`

cburgdorf:
Alright :+1:  Updated it!

pipermerriam:
Eh, would like a test for this to show it works.

cburgdorf:
@pipermerriam It looks like with the recent changes in master, there is no need anymore to have `subscribe` be `async` which means we don't actually need `subscribe_nowait`. Instead, we *could* just drop the `async` from `subscribe` again to turn `async def subscribe` into just `def subscribe`.

The only reason that I can think of why we would still want both `subscribe` / `subscribe_nowait` would be if it is needed for the trio implementation because things there may require `subscribe` to be `async`.

I still updated this PR but it is now turned on its head with `def subscribe_nowait` being the default implementation and `async subscribe` just being an `async` function that delegates to `def subscribe_nowait`.

cburgdorf:
@pipermerriam I opened https://github.com/ethereum/lahja/pull/118 to demonstrate just dropping the `async`. If there is no strong reason why we need an `async def subscribe` I would be in favor of going with #118 instead of this PR. This would essentially keep the `subscribe` API unchanged from the latest stable lahja release (with the exception of it being able to accept coroutines as handlers).

cburgdorf:
Closing this in favor of #118

PR #107, opened 2019-05-29
Ensure full API is represented on EndpointAPI

pipermerriam:
## What was wrong?

The `BaseEndpoint` API isn't properly represented on the `EndpointAPI` meaning that 3rd party libraries that generically specify `EndpointAPI ` as an object type will get type checking errors if they use any of the subscription methods.

## How was it fixed?

Added all of them as `abstractmethod`.

#### Cute Animal Picture

Cute Baby Hippo (14)

### DIFF
--- a/lahja/base.py
+++ b/lahja/base.py
@@ -210,7 +210,9 @@ class EndpointAPI(ABC):
         self
     ) -> Tuple[Tuple[Optional[str], Set[Type[BaseEvent]]], ...]:
         """
-        Return all connected endpoints and their event type subscriptions to this endpoint.
+        Return 2-tuples for all all connected endpoints containing the name of
+        the endpoint (which might be ``None``) coupled with the set of messages
+        the endpoint subscribes to
         """
         ...
 
--- a/lahja/base.py
+++ b/lahja/base.py
@@ -122,15 +122,6 @@ class EndpointAPI(ABC):
         """
         ...
 
-    @abstractmethod
-    def get_connected_endpoints_and_subscriptions(
-        self
-    ) -> Tuple[Tuple[Optional[str], Set[Type[BaseEvent]]], ...]:
-        """
-        Return all connected endpoints and their event type subscriptions to this endpoint.
-        """
-        ...
-
     #
     # Event API
     #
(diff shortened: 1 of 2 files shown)


lahja/base.py
```diff
@@ -211,6 +202,72 @@ def broadcast_nowait(
...
 
+    #
+    # Subscription API
+    #
+    @abstractmethod
+    def get_connected_endpoints_and_subscriptions(
+        self
+    ) -> Tuple[Tuple[Optional[str], Set[Type[BaseEvent]]], ...]:
+        """
+        Return all connected endpoints and their event type subscriptions to this endpoint.
```
lithp: nitpick: it's not clear what it means for a subscription to be to an endpoint. Maybe: "Return all connected endpoints and the set messages each connected endpoint is subscribed to"

lahja/base.py
```diff
@@ -211,6 +202,72 @@ def broadcast_nowait(
...
+        """
+        Block until any other remote endpoint has subscribed to the specified event type
+        from this endpoint.
+        """
+        ...
+
+    @abstractmethod
+    async def wait_until_all_remotes_subscribed_to(
+        self, event: Type[BaseEvent]
+    ) -> None:
```
lithp: nitpick: Not part of this PR but these methods would be slightly easier to use if they included timeouts, you probably never want to call this method without wrapping it in something like [`wait_for`](https://docs.python.org/3/library/asyncio-task.html#asyncio.wait_for).

pipermerriam: @lithp I've been coming around to the `trio` way of doing things which is that `timeout` parameters should not be part of the API but a responsibility of the caller.  

4th paragraph here: https://trio.readthedocs.io/en/latest/reference-core.html#handling-cancellation

I know that this concept doesn't map perfectly onto `asyncio` since asyncio doesn't have as nice of timeout support... but I still think there might be a nugget of wisdom in there.

lithp: Nice link, this makes me even more excited to try working with trio. I'm definitely buying the argument that timeouts are more like exceptions; the code in the best position to determine the right timeout or exception handling strategy is usually relatively far away from the code which might generate those timeouts or exceptions.

PR #108, opened 2019-05-30
Generic wait for subscriptions and connection change

pipermerriam:
Builds from #107 

## What was wrong?

I didn't like that the generic implemetnations of the various subscription based `wait` APIs prescribed implementation details, primarily, that each implementation was required to implement something that streamed an event over the local bus to signal the change.

We also need an api for waiting until a connection is established to a specific endpoint.

## How was it fixed?

Now there are two methods that each endpoint must implement:

- `wait_until_remote_subscriptions_change`
- `wait_until_connections_change`

From these two APIs, all of the following can be constructed generically in the `BaseEndpoint` class.
- `wait_until_connections_change`
    - `wait_until_connected_to` (new API)
- `wait_until_remote_subscriptions_change`
    - `are_all_remotes_subscribed_to`
    - `wait_until_any_remote_subscribed_to`
    - `wait_until_all_remotes_subscribed_to`

#### Cute Animal Picture

Cute Baby Hippo (4)

### DIFF
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -49,7 +49,6 @@ from lahja.common import (
     ConnectionConfig,
     Message,
     Msg,
-    RemoteSubscriptionChanged,
     RequestIDGenerator,
     Subscription,
     SubscriptionsAck,
@@ -276,6 +275,8 @@ class AsyncioEndpoint(BaseEndpoint):
 
     _ipc_path: Path
 
+    _remote_connections_changed: asyncio.Condition
+
     _receiving_queue: "asyncio.Queue[Tuple[Union[bytes, BaseEvent], Optional[BroadcastConfig]]]"
     _receiving_loop_running: asyncio.Event
 
@@ -299,6 +300,9 @@ class AsyncioEndpoint(BaseEndpoint):
             raise Exception(
                 f"TODO: Invalid endpoint name: '{name}'. Must be ASCII encodable string"
             )
+
+        # Signal when a new remote connection is established
+        self._remote_connections_changed = asyncio.Condition()
 
         # storage containers for inbound and outbound connections to other
         # endpoints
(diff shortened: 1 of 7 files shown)


pipermerriam:
@lithp this implements the core of #86 in a less controversial way.  Think that the outbound notifications still need to be cleaned up which is the part that had some controversy but I'll do that in a separate PR.

lahja/asyncio/endpoint.py
```diff
@@ -532,37 +535,56 @@ def connect_to_endpoints_nowait(self, *endpoints: ConnectionConfig) -> None:
...
-                RemoteSubscriptionChanged(), BroadcastConfig(internal=True)
-            )
+        _, pending = await asyncio.wait(
+            (wait_connection_changed,) + wait_remote_subscriptions_change,
+            return_when=asyncio.FIRST_COMPLETED,
+        )
+
+        # We cancel the other pending waits
+        for task in pending:
+            task.cancel()
```
lithp: I can't believe it's so hard to do this in asyncio, it feels like this should be a common pattern but I remember ending up at the same place.

lahja/base.py
```diff
@@ -206,11 +225,9 @@ def broadcast_nowait(
     # Subscription API
     #
     @abstractmethod
-    def get_connected_endpoints_and_subscriptions(
-        self
-    ) -> Tuple[Tuple[Optional[str], Set[Type[BaseEvent]]], ...]:
+    async def wait_until_remote_subscriptions_change(self) -> None:
         """
-        Return all connected endpoints and their event type subscriptions to this endpoint.
+        Block until any subscription change occurs on any remote endpoint.
```
lithp: nitpick: Maybe this comment could also note that this method will also return when the set of remote endpoints changes.

lahja/asyncio/endpoint.py
pipermerriam: If it's any solitude, `trio` doesn't have any equivalent to `asyncio.gather`.

You can see my ugly stab at implementing it here: https://github.com/ethereum/trinity/pull/615/files#diff-32f4c7ca496f37029fb4feaa33835db1R38

tests/core/asyncio/test_asyncio_subscriptions_api.py
```diff
@@ -194,7 +194,7 @@ def noop(event):
     asyncio.ensure_future(server_a.subscribe(WaitSubscription, noop))
 
     await asyncio.wait_for(
-        client.wait_until_any_remote_subscribed_to(WaitSubscription), timeout=0.1
+        client.wait_until_any_remote_subscribed_to(WaitSubscription), timeout=1
```
cburgdorf: Out of curiosity: Any particular reason why this increase is needed?

pipermerriam: I think it just snuck in from some debugging changes.  Should be reverted.

PR #110, opened 2019-05-30
Cleanup local api for sending subscription updates to remotes

### LINKED ISSUE #109: More efficient subscription updates.
## What was wrong?

Right now, the way most of the APIs that do things like update subscriptions operate, we incur some inefficiencies when multiple updates occur in quick succession.  Today, if we were to call `Endpoint.subscribe` for multiple different events in quick succession, we would issue the same number of subscription updates to the remote.

In reality, we should only ever issue the *last* one.

## How can it be fixed?

We should be able to roll something simple with an `Event` and maybe a `Lock` to make it so the call sites that notify of the change call `Event.set`, which is a noop anytime the event has already been set.  Then, once all of the *waits* have been processed, we do something like replace the `Event` with a new fresh one.

The end result goal here is that in the event that multiple calls have been made to *notify* that something has happened, that we only ever issue a single update for all of the stacked up notifications.

pipermerriam:
Builds from #108 

Fixes #109 

## What was wrong?

The manner in which subscription updates were issued had two issues that could be improved.

1. Multiple quick updates to the local subscription API would result in multiple updates being sent, when really only the last update is relevant.
2. The API was only callable from async contexts, requiring the use of `asyncio.ensure_future` when updates occur from a sync context.

## How was it fixed?

Use an `asyncio.Event` that is cycled out each time it's triggered, allowing multiple call sites to call `Event.set()` but only trigger a single update.  This has the bonus benefit of not requiring an async context since `Event.set` is synchronous.

#### Cute Animal Picture

cute-baby-hippos-134-59086cdb9f63f__700

### DIFF
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -449,7 +449,7 @@ class AsyncioEndpoint(BaseEndpoint):
     async def _monitor_subscription_changes(self) -> None:
         self._subscriptions_changed = asyncio.Event()
 
-        while True:
+        while self.is_running:
             # We wait for the event to change and then immediately replace it
             # with a new event.  This **must** occur before any additional
             # `await` calls to ensure that any *new* changes to the
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -446,7 +446,7 @@ class AsyncioEndpoint(BaseEndpoint):
         Tell all inbound connections of our new subscriptions
         """
         # make a copy so that the set doesn't change while we iterate over it
-        subscribed_events = self.subscribed_events
+        subscribed_events = frozenset(self.subscribed_events)
         for remote in self._half_connections.copy():
             await remote.notify_subscriptions_updated(subscribed_events)
         for remote in tuple(self._full_connections.values()):
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -291,6 +291,8 @@ class AsyncioEndpoint(BaseEndpoint):
     _loop: Optional[asyncio.AbstractEventLoop] = None
     _get_request_id: Iterator[RequestID]
 
+    _subscriptions_changed: asyncio.Event
+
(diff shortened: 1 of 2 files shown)


pipermerriam:
@lithp thank you for challenging me on #86 

This implementation is *sooooooo* much simpler and accomplishes everything I wanted.

lahja/asyncio/endpoint.py
```diff
@@ -429,16 +434,25 @@ def subscribed_events(self) -> Set[Type[BaseEvent]]:
...
+            # with a new event.  This **must** occur before any additional
+            # `await` calls to ensure that any *new* changes to the
+            # subscriptions end up operating on the *new* event and will be
+            # picked up in the next iteration of the loop.
+            await self._subscriptions_changed.wait()
+            self._subscriptions_changed = asyncio.Event()
+
+            # make a copy so that the set doesn't change while we iterate
+            # over it
+            subscribed_events = self.subscribed_events
```
lithp: minor: I forgot this was a property and confused myself. Something like `subscribed_events = frozenset(self.subscribed_events)` would have saved me even though it's technically redundant. It's also slightly safer, the current code requires that `subscribed_events` is a property but does nothing to enforce that assumption.

pipermerriam:
@lithp I've really wanted to figure out a way to make `broadcast/request/stream/wait_for`.  I'll open an issues to clean all this up.

lahja/base.py
```diff
@@ -232,8 +232,12 @@ def broadcast_nowait(
     @abstractmethod
     async def wait_until_remote_subscriptions_change(self) -> None:
         """
+<<<<<<< HEAD
         Block until any subscription change occurs on any remote endpoint or
         the set of remote endpoints changes
+=======
+        Block until any subscription change occurs on any remote endpoint.
+>>>>>>> Generic API for waiting for subscription and connection changes
```
lithp: You probably didn't intend to commit this :sweat_smile:

pipermerriam: It is removed in #88

PR #111, opened 2019-05-30
Misc cleanups

pipermerriam:
## What was wrong?

I ditched a branch but it had some cleanups I wanted to maintain that are a bit random.

## How was it fixed?

Combined them into this single PR.

- `__str__` and `__repr__` methods for `RemoteEndpoint`
- some code comments
- a minor bugfix in the resource locking of `RemoteEndpoint.subscribed_messages`
- `__str__` and `__repr__` methods for `AsyncioEndpoint`
- Nicer killing of the main driver process in the benchmarks.
- Wait until later to block on subscription in the benchmark reporter process
- A pre-assertion for a test

#### Cute Animal Picture

IRBSY7C

### DIFF
--- a/lahja/tools/benchmark/process.py
+++ b/lahja/tools/benchmark/process.py
@@ -2,6 +2,8 @@ from abc import ABC, abstractmethod
 import itertools
 import logging
 import multiprocessing
+import os
+import signal
 import time
 from typing import Any, AsyncGenerator, List, NamedTuple, Optional, Tuple  # noqa: F401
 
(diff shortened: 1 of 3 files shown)


lahja/tools/benchmark/process.py
```diff
@@ -42,15 +44,26 @@ def start(self) -> None:
 
     def stop(self) -> None:
```
lithp: This is great! multiprocessing should have this built in rather than defaulting to `SIGTERM` :cry:

pipermerriam:
Looks like another hang bug exposed by the benchmark suite with something causing *mysterious* timeouts (in the request/response part)

PR #116, opened 2019-06-03
Collapse duplicate tests

pipermerriam:
## What was wrong?

There were some duplicate tests for the same functionality.

## How was it fixed?

Expanded one suite of tests to cover the one uncovered case and removed the other suite of tests.  

> I chose the ones from `test_asyncio_subscriptions_api.py` because I think they are slightly more legible.  I am of coure biased since I wrote them...

#### Cute Animal Picture

work-it


pipermerriam:
ping @lithp or @cburgdorf

tests/core/asyncio/test_subscription_wait_apis.py
```diff
@@ -1,90 +0,0 @@
...
-    subscriptions = endpoint3.get_connected_endpoints_and_subscriptions()
-
-    assert (endpoint1.name, {DummyRequestPair}) in subscriptions
-
-    assert endpoint3.is_remote_subscribed_to(endpoint1.name, DummyRequestPair)
-    assert not endpoint3.is_remote_subscribed_to(endpoint2.name, DummyRequestPair)
-    assert not endpoint3.is_remote_subscribed_to("nonexistent", DummyRequestPair)
-
-    assert not endpoint3.is_remote_subscribed_to(endpoint1.name, DummyResponse)
-    assert not endpoint3.is_remote_subscribed_to(endpoint2.name, DummyResponse)
```
lithp: The `test_asyncio_wait_until_any_remote_subscribed_to` test would be a more thorough replacement of this test if it included a `assert client.is_any_remote_subscribed_to(WaitSubscription)` at the end. Currently it's testing that the relevant call doesn't hang forever, `return` would pass that test but it wouldn't pass this one.

pipermerriam:
Ended up adding an assertion like that to all three of the tests.  Good addition.

PR #120, opened 2019-06-06
Ensure local use of endpoint works

pipermerriam:
## What was wrong?

We need the ability for an endpoint to consume events from itself (as if it were connected).

## How was it fixed?

This is a *partial* fix since it's unclear what to do about the varios `wait_until` and `is_connected` APIs.

This ensures that all of the primary event base APIs on an endpoint work as expected for local consumption of events.

#### Cute Animal Picture

6f7f7d937161bd52f5638bd7ed9bc8ed

### DIFF
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -52,6 +52,7 @@ from lahja.common import (
     Msg,
     RequestIDGenerator,
     Subscription,
+    should_endpoint_receive_item,
 )
 from lahja.exceptions import (
     ConnectionAttemptRejected,
@@ -388,7 +389,11 @@ class AsyncioEndpoint(BaseEndpoint):
         seen: Set[str] = set()
 
         for config in endpoints:
-            if config.name in seen:
+            if config.name == self.name:
+                raise ConnectionAttemptRejected(
+                    f"Cannot connect an endpoint to itself."
+                )
+            elif config.name in seen:
                 raise ConnectionAttemptRejected(
                     f"Trying to connect to {config.name} twice. Names must be uniqe."
                 )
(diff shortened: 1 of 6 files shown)


lahja/asyncio/endpoint.py
```diff
@@ -612,10 +616,11 @@ def stop(self) -> None:
     ) -> None:
         item.bind(self, id)
 
+        await self._process_item(item, config)
```
cburgdorf: I think this is ok for this PR but we might want to reconsider this. The implication of this is that every broadcasted event (some are high frequency) is going into `_process_item` which doesn't do much if the endpoint isn't subscribed itself but it does do *something* 

I *think* we should probably only process it if we are actually subscribed to the event (similar handling as if it was an external event that we are subscribed for)

lahja/asyncio/endpoint.py
```diff
@@ -388,7 +388,11 @@ def _throw_if_already_connected(self, *endpoints: ConnectionConfig) -> None:
         seen: Set[str] = set()
 
         for config in endpoints:
-            if config.name in seen:
+            if config.name == self.name:
+                raise ConnectionAttemptRejected(
+                    f"Trying to connect to {config.name} twice. Names must be uniqe."
```
cburgdorf: This should probably be a different message. E.g. "Can not connect to own endpoint"

lahja/asyncio/endpoint.py
```diff
@@ -612,10 +616,11 @@ def stop(self) -> None:
     ) -> None:
         item.bind(self, id)
 
+        await self._process_item(item, config)
+
         if config is not None and config.internal:
             # Internal events simply bypass going over the event bus and are
             # processed immediately.
```
cburgdorf: This comment could be changed to something like:

>No further processing, internal events aren't leaving the local endpoint 

(you get the idea)

PR #121, opened 2019-06-06
Flesh out self-wait-subscribe APIs

pipermerriam:
## What was wrong?

I got in a hurry and didn't take into account that this needed to be sorted out.

The various APIs for *waiting* on subscriptions, checking subscriptions, waiting on connections and checking connections don't always behave in intuitive ways when dealing with a singular endpoint as if it was two endpoints which is a feature that we want to support.

## How was it fixed?

I don't think there was *one* correct behavior here, at least not by default.  Before the revers connections change, you did have to manually connect an endpoint to itself for this to work.  This is now replaced by a call to `EndpointAPI.enable_local_broadcast()`

Once this has been enabled, the endpoint will treat itself exactly as if it was another connected remote.  This includes triggering all of the correct `wait_` apis.

Comparing this to the previous state of affairs prior to reverse connections, I believe it is functionally equivalent. In previous code, if you simply replaced `endpoint.connect_to_endpoint(config_for_self)` with `endpoint.enable_local_broadcast()` I believe that the functionality would be identical.

#### Cute Animal Picture

PKTVTD0

### DIFF
--- a/lahja/base.py
+++ b/lahja/base.py
@@ -305,11 +305,6 @@ class EndpointAPI(ABC):
 
     name: str
 
-    _remote_connections_changed: ConditionAPI
-    _remote_subscriptions_changed: ConditionAPI
-
-    _connections: Set[RemoteEndpointAPI]
-
     @property
     @abstractmethod
     def is_running(self) -> bool:
@@ -392,6 +387,20 @@ class EndpointAPI(ABC):
     async def wait_until_connections_change(self) -> None:
         """
         Block until the set of connected remote endpoints changes.
+        """
+        ...
+
+    @property
+    @abstractmethod
+    def is_local_broadcast_enabled(self) -> bool:
+        ...
+
+    @abstractmethod
+    def enable_local_broadcast(self) -> None:
+        """
+        When enabled, this endpoint will treat itself as one of it's connected
+        remote endpoints when dealing with event broadcasting and
+        subscriptions.
         """
         ...
 
@@ -540,13 +549,31 @@ class EndpointAPI(ABC):
         """
         ...
 
+    @abstractmethod
+    def get_subscribed_events(self) -> Set[Type[BaseEvent]]:
+        ...
+
 
 class BaseEndpoint(EndpointAPI):
     """
     Base class for endpoint implementations that implements shared/common logic
     """
 
+    _local_bradcast_enabled: EventAPI
+
+    _remote_connections_changed: ConditionAPI
+    _remote_subscriptions_changed: ConditionAPI
+
+    _connections: Set[RemoteEndpointAPI]
+
     logger = logging.getLogger("lahja.endpoint.Endpoint")
+
+    def __init__(self, name: str) -> None:
+        self.name = name
+
+        # storage containers for inbound and outbound connections to other
+        # endpoints
+        self._connections = set()
 
     def __str__(self) -> str:
         return f"Endpoint[{self.name}]"
@@ -572,6 +599,8 @@ class BaseEndpoint(EndpointAPI):
             await self.connect_to_endpoint(config)
 
     def is_connected_to(self, endpoint_name: str) -> bool:
+        if self.is_local_broadcast_enabled and endpoint_name == self.name:
+            return True
         return any(endpoint_name == remote.name for remote in self._connections)
 
     async def wait_until_connected_to(self, endpoint_name: str) -> None:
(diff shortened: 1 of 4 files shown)


pipermerriam:
This also means I'll be releasing lahja `0.14.0` with great shame.

pipermerriam:
@cburgdorf and @lithp I sent this to both of you but I'm fine with just one review, though I woudn't mind confirmation that this is an API you like or suggestions on something alternate.

lahja/base.py
```diff
@@ -395,6 +390,20 @@ def get_connected_endpoints_and_subscriptions(
         """
         ...
 
+    @property
+    @abstractmethod
+    def is_local_broadcast_enabled(self) -> bool:
```
cburgdorf: I'm not sold on this API to be honest. If we just look at the history of the project, it is true that in the previous version you had to explicitly connect to your own endpoint to be able to act upon your own events. But looking further back this features has actually existed from day one *without* any explicit API (that changed when we moved away from the central router architecture to the mesh architecture).

But history aside, do we really gain much from making this opt-in? The only pro argument that I see is a (probably tiny) performance improvement for not having to check the subscribed events if the feature is disabled. Other than that, performance should not play a key role here as we don't process local events if we aren't locally subscribed.

I would also argue that it makes the APIs feel less coherent. E.g I'm not sure if `True` is the right answer to `is_connected_to('myself')`. Plus, having it either return `True` or `False` depending on the fact if `local_broadcast` is enabled or not doesn't feel right to me. Nothing changes in terms of "being connected". 

In my mind `is_connected_to('myself')` could simply always throw an exception stating something like *"Endpoints do not hold connections to themselves"* (Relying on you native speakers to round the edges but you get the idea).

I **do want** to be able to say `is_any_remote_subscribed_to(SomeEvent)` and have that take our own subscriptions into account but I think that we should rename that API to `is_any_endpoint_subscribed_to` to make that more coherent. *Any* **endpoint**  does include our self but *any* **remote**  does not sound like an API that should factor in our own subscriptions.

I'm not saying *over my dead body* but I would like to raise these concerns and also hear @lithp opinion on this.

lahja/base.py
```diff
@@ -395,6 +390,20 @@ def get_connected_endpoints_and_subscriptions(
...
+        ...
+
+    @abstractmethod
+    def enable_local_broadcast(self) -> None:
+        """
+        When enabled, this endpoint will treat itself as one of it's connected
+        remote endpoints when dealing with event broadcasting and
+        subscriptions.
+        """
```
cburgdorf: I think this docstring sums up what I don't like about this change: **When enabled** (some core APIs behave very different depending on the fact of the feature is enabled/disabled) (...) *treat itself as one of it's connected remote endpoints* -> To me feels like some APIs are just telling lies in subtle ways. *Maybe* some of this could be changed with API renames though (e.g. `is_any_remote_subscribed_to` -> `is_any_endpoint_subscribed_to`)

cburgdorf:
@pipermerriam Taking your comment from discord here:

>I do believe the API should be opt-in (it was beforehand as well)  Without this, `endpoint.are_all_endpoints_subscribed_to(...)` has really un-ergonomic and surprising behavior (that it won't return true until the local endpoint also subscribes which is rarely what the caller wants in a situation that spans processes and is truly a broadcast to consumers)

I think I would rather resolve that with an extra parameter `exclude_own_endpoint` that defaults to `True`.

So, while I'm all for updating namof the APIs and ergonomics, I'd like to hear an argument for why this functionality is wrong but the previous functionality was right (requring that you establish a self connection over an actual socket).

Well, requiring a connection to self was more like a band aid after the topology switched from centra-router-based to mesh-based. When it was central-router-based, subscribing to events on the same endpoint that created these events just worked out-of-the-box without any special API calls.

lahja/base.py
pipermerriam: I think the two awkward parts for me that keep me from being immediately :+1: about making this just the default *baseline* behavior are `is_connected_to` and `are_all_remotes_subscribed` to.

`is_connected_to(<myself>)` returning `True` unconditionally could be fine, just needs to be well documented.  Not a big hang-up.  That naturally must then extend to `get_connected_remotes_and_subscriptions` including `<myself>` which is a bit more weird, but again, not that big a deal.

The big one for me is `are_all_remotes_subscribed_to(...)`.  I feel like it is very un-ergonomic to require an endpoint to subscribe to the thing itself for this to return true, since it *seems* like the common case is some central coordinator that is waiting for everyone else to subscribe.  **BUT** I'm not sure that use case is actually real.  In `trinity` I think that we'd need the *self* endpoint to also be subscribed so maybe my arguments are moot..... (maybe I just talked myself out of it).

Curious also to hear @lithp opinion and I think I'm *"ok"* with just making this the default behavior.

pipermerriam:
Closed in favor of #122

PR #122, opened 2019-06-07
Self subscribe API with auto-opt-in

pipermerriam:
## What was wrong?

I got in a hurry and didn't take into account that this needed to be sorted out.

The various APIs for *waiting* on subscriptions, checking subscriptions, waiting on connections and checking connections don't always behave in intuitive ways when dealing with a singular endpoint as if it was two endpoints which is a feature that we want to support.

## How was it fixed?

Endpoints now treat themselves as a connected endpoint.  All of the various *wait* APIs will treat *self* as part of the set of connected endpoints.

This also results in renaming some of the APIs that previously used the term *remote* to instead use *endpoint* since it is inconsistent to refer to ourselves as a remote.

The only exception is the `wait_until_all_endpoints_subscribed` and `are_all_endgoints_subscribed` which each take an optional extra parameter `exclude_self` which defaults to `False`.

#### Cute Animal Picture

PKTVTD0

### DIFF
--- a/lahja/base.py
+++ b/lahja/base.py
@@ -103,15 +103,14 @@ class RemoteEndpointAPI(ABC):
     async def stop(self) -> None:
         ...
 
-    @abstractmethod
-    async def _run(self) -> None:
-        ...
-
     #
     # Core external API
     #
     @abstractmethod
     def get_subscribed_events(self) -> Set[Type[BaseEvent]]:
+        """
+        Return the event types this endpoint is current subscribed to.
+        """
         ...
 
     @abstractmethod
@@ -133,6 +132,10 @@ class RemoteEndpointAPI(ABC):
 
     @abstractmethod
     async def wait_until_subscription_initialized(self) -> None:
+        """
+        Block until at least one `SubscriptionsUpdated` event has been
+        received.
+        """
         ...
 
 
@@ -305,11 +308,6 @@ class EndpointAPI(ABC):
 
     name: str
 
-    _remote_connections_changed: ConditionAPI
-    _remote_subscriptions_changed: ConditionAPI
-
-    _connections: Set[RemoteEndpointAPI]
-
     @property
     @abstractmethod
     def is_running(self) -> bool:
@@ -479,7 +477,7 @@ class EndpointAPI(ABC):
     # Subscription API
     #
     @abstractmethod
-    async def wait_until_remote_subscriptions_change(self) -> None:
+    async def wait_until_endpoint_subscriptions_change(self) -> None:
         """
         Block until any subscription change occurs on any remote endpoint or
         the set of remote endpoints changes
@@ -487,7 +485,7 @@ class EndpointAPI(ABC):
         ...
 
     @abstractmethod
-    def is_remote_subscribed_to(
+    def is_endpoint_subscribed_to(
         self, remote_endpoint: str, event_type: Type[BaseEvent]
     ) -> bool:
         """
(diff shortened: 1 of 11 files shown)


pipermerriam:
This is the same as #121 but without it always being on.  This approach is growing on me.

lahja/base.py
```diff
@@ -540,14 +535,30 @@ def are_all_remotes_subscribed_to(self, event_type: Type[BaseEvent]) -> bool:
         """
         ...
 
+    @abstractmethod
+    def get_subscribed_events(self) -> Set[Type[BaseEvent]]:
+        ...
```
cburgdorf: It think this should also get a docstring

tests/core/asyncio/test_asyncio_subscriptions_api.py
```diff
@@ -225,11 +225,16 @@ def noop(event):
...
     server_c.subscribe(WaitSubscription, noop)
+    await asyncio.sleep(0.01)
     assert got_subscription.is_set() is False
     server_a.subscribe(WaitSubscription, noop)
+    await asyncio.sleep(0.01)
     assert got_subscription.is_set() is False
     server_b.subscribe(WaitSubscription, noop)
     await asyncio.sleep(0.01)
+    assert got_subscription.is_set() is False
+    client.subscribe(WaitSubscription, noop)
```
cburgdorf: What are your thoughts on giving `wait_until_all_remotes_subscribed_to` an extra parameter `exclude_self` to support the case where you want to use that API but **without** waiting for your own subscription?

tests/core/asyncio/test_broadcast_config.py
```diff
@@ -18,7 +18,9 @@ class BroadcastEvent(BaseEvent):
     client_a.subscribe(BroadcastEvent, return_queue.put_nowait)
     client_b.subscribe(BroadcastEvent, return_queue.put_nowait)
 
-    await server.wait_until_all_remotes_subscribed_to(BroadcastEvent)
```
cburgdorf: E.g. this test could then stay as it was before but use

`await server.wait_until_all_remotes_subscribed_to(BroadcastEvent, exclude_self=True)`

tests/core/asyncio/test_broadcast_config.py
```diff
@@ -18,7 +18,9 @@ class BroadcastEvent(BaseEvent):
...
+    await server.wait_until_remote_subscribed_to(client_a.name, BroadcastEvent)
+    await server.wait_until_remote_subscribed_to(client_b.name, BroadcastEvent)
```
cburgdorf: Oh and I think we should really do the rename`wait_until_remote_subscribed_to` -> `wait_until_endpoint_subscribed_to`. The name `remote` is a bit misleading here.

pipermerriam:
Waiting till monday (actually Tuesday) to merge and release this.  Doing cleanup now.  If anyone wants to merge and release on monday when I'll be traveling feel free.

lithp:
Look like Christoph beat me to it but I like this, and the `exclude_self` parameter is a great idea :+1:

pipermerriam:
Good to release on Monday as `0.14.0` and @carver you can develop against `master` now if you'd like to get a head-start.

carver:
How about `include_self=True` as the argument so that the variable isn't a negative. Negative bool flags lead to double negative checks that are more error prone, like: `if not exclude_self:`.

cburgdorf:
Good point @carver I've never thought about that :+1:

cburgdorf:
@carver https://github.com/ethereum/lahja/pull/123

PR #123, opened 2019-06-11
Change `exclude_self` parm to `include_self`

cburgdorf:
## What was wrong?

Negative bool flags lead to double negative checks
that are more error prone, like: `if not exclude_self`

## How was it fixed?

Changed `exclude_self` param that defaulted to `False` to `include_self` param that defaults to `True`.

#### Cute Animal Picture

### DIFF
--- a/lahja/base.py
+++ b/lahja/base.py
@@ -532,7 +532,7 @@ class EndpointAPI(ABC):
 
     @abstractmethod
     async def wait_until_all_endpoints_subscribed_to(
-        self, event: Type[BaseEvent], *, exclude_self: bool = False
+        self, event: Type[BaseEvent], *, include_self: bool = True
     ) -> None:
         """
         Block until all currently connected remote endpoints are subscribed to the specified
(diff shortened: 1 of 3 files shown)


cburgdorf:
I'm holding off releasing `0.14.0` to get that last bit in to avoid this becoming a breaking change with the next release.

.pytest_cache/v/cache/nodeids
```diff
@@ -0,0 +1,58 @@
+[
+  "tests/core/asyncio/test_asyncio_subscriptions_api.py::test_asyncio_stream_api_updates_subscriptions[client-server]",
+  "tests/core/asyncio/test_asyncio_subscriptions_api.py::test_asyncio_stream_api_updates_subscriptions[server-client]",
```
pipermerriam: Looks like this missed the `gitignore`.

cburgdorf: Oops https://github.com/ethereum/lahja/pull/125

PR #126, opened 2019-06-11
Trio base endpoint implementation

pipermerriam:
replaces #56

built on #133 #135 #136

## What was wrong?

We need a trio based endpoint implementation.

## How was it fixed?

Wrote an implementation using the `trio` library.

#### Cute Animal Picture

4zvN5a

### DIFF
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -51,7 +51,6 @@ from lahja.common import (
     ConnectionConfig,
     Message,
     Msg,
-    RequestIDGenerator,
     Subscription,
     should_endpoint_receive_item,
 )
@@ -199,19 +198,11 @@ class AsyncioEndpoint(BaseEndpoint):
     _sync_handler: DefaultDict[Type[BaseEvent], List[SubscriptionSyncHandler]]
 
     _loop: Optional[asyncio.AbstractEventLoop] = None
-    _get_request_id: Iterator[RequestID]
 
     _subscriptions_changed: asyncio.Event
 
     def __init__(self, name: str) -> None:
         super().__init__(name)
-
-        try:
-            self._get_request_id = RequestIDGenerator(name.encode("ascii") + b":")
-        except UnicodeDecodeError:
-            raise Exception(
-                f"TODO: Invalid endpoint name: '{name}'. Must be ASCII encodable string"
-            )
 
         # Signal when a new remote connection is established
         self._remote_connections_changed = asyncio.Condition()  # type: ignore
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -153,7 +153,7 @@ class AsyncioRemoteEndpoint(BaseRemoteEndpoint):
     #
     # Running API
     #
-    @asynccontextmanager  # type: ignore
+    @asynccontextmanager
     async def run(self) -> AsyncIterator[RemoteEndpointAPI]:
         await self._start()
 
@@ -284,7 +284,7 @@ class AsyncioEndpoint(BaseEndpoint):
     #
     # Running API
     #
-    @asynccontextmanager  # type: ignore
+    @asynccontextmanager
     async def run(self) -> AsyncIterator[EndpointAPI]:
         if not self._loop:
             self._loop = asyncio.get_event_loop()
@@ -421,7 +421,7 @@ class AsyncioEndpoint(BaseEndpoint):
     # Server API
     #
     @classmethod
-    @asynccontextmanager  # type: ignore
+    @asynccontextmanager
     async def serve(cls, config: ConnectionConfig) -> AsyncIterator["AsyncioEndpoint"]:
         endpoint = cls(config.name)
         async with endpoint.run():
--- a/lahja/base.py
+++ b/lahja/base.py
@@ -10,6 +10,7 @@ from typing import (  # noqa: F401
     Callable,
     Dict,
     Iterable,
+    Iterator,
     List,
     NamedTuple,
     Optional,
@@ -31,12 +32,13 @@ from .common import (
     Hello,
     Message,
     Msg,
+    RequestIDGenerator,
     Subscription,
     SubscriptionsAck,
     SubscriptionsUpdated,
 )
 from .exceptions import ConnectionAttemptRejected, RemoteDisconnected
-from .typing import ConditionAPI, EventAPI, LockAPI
+from .typing import ConditionAPI, EventAPI, LockAPI, RequestID
 
 TResponse = TypeVar("TResponse", bound=BaseEvent)
 TWaitForEvent = TypeVar("TWaitForEvent", bound=BaseEvent)
@@ -587,10 +589,19 @@ class BaseEndpoint(EndpointAPI):
 
     _connections: Set[RemoteEndpointAPI]
 
+    _get_request_id: Iterator[RequestID]
+
     logger = logging.getLogger("lahja.endpoint.Endpoint")
 
     def __init__(self, name: str) -> None:
         self.name = name
+
+        try:
+            self._get_request_id = RequestIDGenerator(name.encode("ascii") + b":")
+        except UnicodeDecodeError:
+            raise Exception(
+                f"TODO: Invalid endpoint name: '{name}'. Must be ASCII encodable string"
+            )
 
         # storage containers for inbound and outbound connections to other
         # endpoints
--- a/lahja/base.py
+++ b/lahja/base.py
@@ -241,11 +241,15 @@ class BaseRemoteEndpoint(RemoteEndpointAPI):
     async def _process_incoming_messages(self) -> None:
         self._running.set()
 
-        # Send the hello message
-        await self.send_message(Hello(self._local_name))
+        try:
+            # Send the hello message
+            await self.send_message(Hello(self._local_name))
+            # Wait for the other endpoint to identify itself.
+            hello = await self.conn.read_message()
+        except RemoteDisconnected:
+            self._stopped.set()
+            return
 
-        # Wait for the other endpoint to identify itself.
-        hello = await self.conn.read_message()
         if isinstance(hello, Hello):
             self._name = hello.name
             self.logger.debug(
--- a/lahja/trio/endpoint.py
+++ b/lahja/trio/endpoint.py
@@ -1,0 +1,807 @@
+import asyncio
+import collections
+import itertools
+import logging
+import pathlib
+import pickle
+from typing import (
+    Any,
+    AsyncGenerator,
+    AsyncIterator,
+    Awaitable,
+    Callable,
+    DefaultDict,
+    Dict,
+    List,
+    Optional,
+    Set,
+    Tuple,
+    Type,
+    TypeVar,
+    Union,
+    cast,
+)
+
+from async_generator import asynccontextmanager
+import trio
+import trio_typing
+
+from lahja import constants
+from lahja.base import (
+    BaseEndpoint,
+    BaseRemoteEndpoint,
+    ConnectionAPI,
+    EndpointAPI,
+    RemoteEndpointAPI,
+    TResponse,
+    TStreamEvent,
+    TSubscribeEvent,
+)
+from lahja.common import (
+    BaseEvent,
+    BaseRequestResponseEvent,
+    Broadcast,
+    BroadcastConfig,
+    ConnectionConfig,
+    Message,
+    Msg,
+    Subscription,
+    should_endpoint_receive_item,
+)
+from lahja.exceptions import (
+    ConnectionAttemptRejected,
+    LifecycleError,
+    RemoteDisconnected,
+    UnexpectedResponse,
+)
+from lahja.typing import ConditionAPI, RequestID
+
+from .future import Future
+
+
+class TrioConnection(ConnectionAPI):
+    logger = logging.getLogger("lahja.trio.Connection")
+
+    def __init__(self, socket: trio.SocketStream) -> None:
+        self._socket = socket
+        self._msg_send_channel, self._msg_receive_channel = cast(
+            Tuple[trio.abc.SendChannel[Message], trio.abc.ReceiveChannel[Message]],
+            trio.open_memory_channel(100),
+        )
+        self._write_lock = trio.Lock()
+        super().__init__()
+
+    def __str__(self) -> str:
+        return f"TrioConnection[{self._socket}]"
+
+    def __repr__(self) -> str:
+        return f"<{self}>"
+
+    async def aclose(self) -> None:
+        await self._socket.aclose()
+
+    #
+    # Connection API
+    #
+    @classmethod
+    async def connect_to(cls, path: pathlib.Path) -> "TrioConnection":
+        socket = await trio.open_unix_socket(str(path))
+        cls.logger.debug("Opened connection to %s: %s", path, socket)
+        return cls(socket)
+
+    async def send_message(self, message: Msg) -> None:
+        msg_data = pickle.dumps(message)
+        size = len(msg_data)
+        try:
+            async with self._write_lock:
+                await self._socket.send_all(size.to_bytes(4, "little") + msg_data)
+        except (trio.ClosedResourceError, trio.BrokenResourceError) as err:
+            raise RemoteDisconnected from err
+
+    async def read_message(self) -> Message:
+        buffer = bytearray()
+
+        while len(buffer) < 4:
+            try:
+                data = await self._socket.receive_some(4 - len(buffer))
+            except (trio.ClosedResourceError, trio.BrokenResourceError) as err:
+                raise RemoteDisconnected from err
+
+            if data == b"":
+                raise RemoteDisconnected()
+
+            buffer.extend(data)
+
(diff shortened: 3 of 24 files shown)


lahja/base.py
```diff
@@ -587,11 +593,20 @@ class BaseEndpoint(EndpointAPI):
...
     def __init__(self, name: str) -> None:
         self.name = name
 
+        try:
+            self._get_request_id = RequestIDGenerator(name.encode("ascii") + b":")
+        except UnicodeDecodeError:
+            raise Exception(
+                f"TODO: Invalid endpoint name: '{name}'. Must be ASCII encodable string"
+            )
+
```
pipermerriam: TODO: extract

lahja/base.py
```diff
@@ -221,16 +223,20 @@ def name(self) -> str:
     async def wait_stopped(self) -> None:
         await self._stopped.wait()
 
+    @property
     def is_running(self) -> bool:
-        return not self.is_stopped and self.running.is_set()
+        return not self.is_stopped and self._running.is_set()
```
pipermerriam: TODO: extract

lahja/asyncio/endpoint.py
```diff
@@ -164,7 +163,7 @@ def __init__(
             await self.stop()
 
     async def _start(self) -> None:
-        self._task = asyncio.ensure_future(self._run())
+        self._task = asyncio.ensure_future(self._process_incoming_messages())
```
pipermerriam: TODO: extract

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,913 @@
...
+
+    @property
+    def is_stopped(self) -> bool:
+        return self._stopped.is_set()
+
+    async def wait_started(self) -> None:
+        await self._running.wait()
+
+    async def wait_stopped(self) -> None:
+        await self._stopped.wait()
```
pipermerriam: I spent some time trying to make a `Runnable` abstraction but it ended up making things feel really disconnected complex to understand.  Maybe worth trying for again once this code has settled.

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,913 @@
...
+RequestResponseChannelPair = Tuple[
+    trio.abc.SendChannel[BaseEvent], trio.abc.ReceiveChannel[BaseEvent]
+]
+
+
+class TrioEndpoint(BaseEndpoint):
+    _stream_channels: Dict[
+        Type[BaseEvent], "weakref.WeakSet[trio.abc.SendChannel[BaseEvent]]"
+    ]
+    _pending_requests: Dict[RequestID, "trio.abc.SendChannel[BaseEvent]"]
```
pipermerriam: TODO: don't need quotes for this type hint

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,913 @@
...
+        async with trio.open_nursery() as nursery:
+            for config in endpoints:
+                nursery.start_soon(self.connect_to_endpoint, config)
+
+    async def connect_to_endpoint(self, config: ConnectionConfig) -> None:
+        """
+        Establish a connection to a named endpoint server over an IPC socket.
+        """
+        if not self.is_running:
+            raise Exception(
```
pipermerriam: TODO: proper exception class

pipermerriam:
This is ready for review but only the latest commit is relevant.  The others are part of other pull requests.

pipermerriam:
I won't be surprised if there are a few underlying issues but this should have rough parity to the existing asyncio implementation and I'd like to start using it as soon as possible.  One thing that I want to focus on a little is unification of the test suites since they are effectively duplicates.

lithp:
About to dive in, 1500 is a lot of lines but I'm excited to see this shipped!

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,905 @@
...
+        with trio.fail_after(2):
+            await self.wait_started()
+
+    async def _stop(self) -> None:
+        if self.is_stopped:
+            return
+        self._stopped.set()
+        await self._cleanup()
+
+    async def _run(self) -> None:
```
lithp: Huh, why did you decide to add this internal buffer, maybe this improved performance in the benchmark? The kernel already adds [a pretty big buffer](https://unix.stackexchange.com/questions/424380/what-values-may-linux-use-for-the-default-unix-socket-buffer-size/424381#424381) here, I was thinking about ways to simplify this code but the easiest way might be to remove it entirely and rely on the buffer that the OS provides.

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,905 @@
...
+    @classmethod
+    async def connect_to(cls, path: pathlib.Path) -> "TrioConnection":
+        socket = await trio.open_unix_socket(str(path))
+        cls.logger.debug("Opened connection to %s: %s", path, socket)
+        return cls(socket)
+
+    async def send_message(self, message: Msg) -> None:
+        msg_data = pickle.dumps(message)
+        size = len(msg_data)
+        async with self._lock:
```
lithp: I guess this lock is because `send_all` can switch away when there's backpressure and then throw an exception if we try to use it reentrantly. I was surprised that trio didn't lock for us but then found [this thread](https://github.com/python-trio/trio/issues/328). 

I think future Brian's will have an easier time reading this if the lock was called something like `_write_lock` and if there to be a comment saying what it's doing, maybe even with a link to the above thread.

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,905 @@
...
+            try:
+                # TODO: look into buffering the writes
```
lithp: Apparently if a cancel happens during `send_all` the socket is corrupted, so a cancel in a task calling `broadcast()` would cause the endpoint to break. Fixing the cancellation problem might involve `send_message` adding the message into a short queue, with another task which reads from the queue and sends, that would also add some buffering for free!

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,905 @@
...
+        async with cast(TrioConnection, self.conn).run():
+            await self._process_incoming_messages()
+
+    async def _start(self, nursery: trio_typing.Nursery) -> None:
+        if self.is_running:
+            raise LifecycleError("RemoteEndpoint is already running")
+        elif self.is_stopped:
+            raise LifecycleError("RemoteEndpoint has already been run and stopped")
+
+        nursery.start_soon(self._run)
```
lithp: It looks like `trio.Nursery.start` is meant to designed to make this pattern easier to write: https://trio.readthedocs.io/en/latest/reference-core.html#trio.Nursery.start

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,905 @@
...
+        async with trio.open_nursery() as nursery:
+            await self._start(nursery)
+            try:
+                yield self
+            finally:
+                await self.stop()
+
+    async def _run(self) -> None:
+        async with cast(TrioConnection, self.conn).run():
+            await self._process_incoming_messages()
```
lithp: I know this goes the other direction from another comment I left, but if `TrioConnection` didn't need to be run and instead read directly from the socket w/o buffering then this cast wouldn't be necessary and some of these methods could be combined:

```python
@asynccontextmanager
async def run(self) -> AsyncIterator[RemoteEndpointAPI]:
  if self.is_running:
    raise LifecycleError("RemoteEndpoint is already running")
  elif self.is_stopped:
    raise LifecycleError("RemoteEndpoint has already been run and stopped")

  async with trio.open_nursery() as nursery:
    nursery.start_soon(self._process_incoming_messages)
    try:
      yield self
    finally:
      await self.stop()
```

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,905 @@
...
+                await send_channel.aclose()
+                break
+
+        # Trigger the subscription change since the event will be removed from
+        # the set once the local reference to the channel goes out of scope.
+        self._subscriptions_changed.set()
+
+    TWaitForEvent = TypeVar("TWaitForEvent", bound=BaseEvent)
+
+    async def wait_for(self, event_type: Type[TWaitForEvent]) -> TWaitForEvent:
```
lithp: I think this overrides and exactly duplicates `lahja.base.BaseEndpoint.wait_for`

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,905 @@
...
+    ) -> AsyncGenerator[TStreamEvent, None]:
+        """
+        Stream all events that match the specified event type. This returns an
+        ``AsyncIterable[BaseEvent]`` which can be consumed through an ``async for`` loop.
+        An optional ``num_events`` parameter can be passed to stop streaming after a maximum amount
+        of events was received.
+        """
+        (send_channel, receive_channel) = cast(
+            StreamChannelPair, trio.open_memory_channel(100)
+        )
```
lithp: Huh, it's annoying that we have to add all these casts everywhere, it might be worth adding our own typestubs, or using someone elses: https://github.com/python-trio/trio-typing

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,905 @@
...
+
+            try:
+                next(counter)
+            except StopIteration:
+                await send_channel.aclose()
+                break
+
+        # Trigger the subscription change since the event will be removed from
+        # the set once the local reference to the channel goes out of scope.
+        self._subscriptions_changed.set()
```
lithp: This feels dangerous. If the weakref *hasn't* been cleaned up, and when it's cleaned up is entirely up to implementation details, then `_monitor_subscription_changes` will check subscriptions and find that the subscription still exists, even though it doesn't. Remote endpoints will never get the message that the subscription is gone until another subscription update happens.

lithp: It looks like `weakref` includes [finalizers](https://docs.python.org/3/library/weakref.html#weakref.finalize), which could be used to notify `_monitor_subscription_changes`

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,905 @@
...
+        :class:`~lahja.common.BroadcastConfig` to decide where the request
+        should be broadcasted to. By default, requests are broadcasted across
+        all connected endpoints with their consuming call sites.
+        """
+        request_id = next(self._get_request_id)
+
+        # Create an asynchronous generator that we use to pipe the result
+        send_channel, receive_channel = cast(
+            RequestResponseChannelPair, trio.open_memory_channel(0)
+        )
```
lithp: I'm surprised that trio doesn't have a `Future`. A channel can do everything a future can do and more, sure, but this is so much more code than a future would have been!

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,905 @@
...
+            # _monitor_subscription_changes
+            #    Monitors an event for local changes to subscriptions to
+            #    propagate those changes to remotes.
+            nursery.start_soon(self._monitor_subscription_changes)
+
+            # wait for the sub-processes to start before marking the endpoint
+            # as running.
+            await self._message_processing_loop_running.wait()
+            await self._connection_loop_running.wait()
+            await self._process_broadcasts_running.wait()
```
lithp: I think all these could be removed with `Nursery.start`

lahja/trio/endpoint.py
pipermerriam: I'll try that.  I'm not entirely sure this is actually necessary so I'll verify the performance gain is real.

lahja/trio/endpoint.py
pipermerriam: :+1:

lahja/trio/endpoint.py
pipermerriam: Yeah, remember that from the docs.  I'll open an issue to track refactoring this to use something like that pattern.

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,905 @@
...
+    async def _process_broadcasts(
+        self,
+        channel: trio.abc.ReceiveChannel[ConcreteBroadcast],
+        nursery: trio_typing.Nursery,
+    ) -> None:
+        self._process_broadcasts_running.set()
+        async for (item, config) in channel:
+            nursery.start_soon(self.broadcast, item, config)
+
+    async def _process_connections(
```
lithp: Why did you decide to pull this out into another coro+channel? If the nursery is added somewhere like `self.nursery` then `connect_to_endpoint` could do the connect and then call `nursery.start_soon`.

lahja/trio/endpoint.py
pipermerriam: :+1:

lahja/trio/endpoint.py
pipermerriam: We're already using `trio-typing`.  I'm not sure it comes with anything for this but I'll look.

lahja/trio/endpoint.py
pipermerriam: :+1:

lahja/trio/endpoint.py
pipermerriam: Yeah, I was surprised as well.  Supposedly it should be pretty easy to write one..

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,905 @@
...
+            return
+        self._stopped.set()
+        await self._cleanup()
+
+    async def _run(self) -> None:
+        # This send/receive channel is fed by connected `RemoteEndpoint`
+        # objects which feed received events into the send side of the channel.
+        # Those messages are then retrieved from the receive channel by the
+        # `_process_received_messages` daemon.
+        (self._message_send_channel, message_receive_channel) = cast(
```
lithp: The fact that trio gives us two channels makes naming hard, but `_message_send_channel` does not sound like the channel we receive messages with, it took me a bit to figure out what it was for. Maybe `_incoming_message_channel_send` and `_incoming_message_channel_receive`?

lahja/trio/endpoint.py
pipermerriam: This might be a leftover from a previous design problem that no-longer exists. Will very and remove if it isn't needed.

lahja/trio/endpoint.py
pipermerriam: I'll see if I can find better names.  Agreed on this being a confusing one.

tests/core/trio/test_trio_endpoint_stream.py
```diff
@@ -0,0 +1,72 @@
...
+class EventTest(BaseEvent):
+    def __init__(self, value):
+        self.value = value
+
+
+@pytest.mark.trio
+async def test_trio_endpoint_stream_without_limit(endpoint_pair):
+    alice, bob = endpoint_pair
+
+    async with trio.open_nursery() as nursery:
```
lithp: This is probably a good fixture candidate

lahja/trio/endpoint.py
lithp: Actually, looking at it again, if they're waiting on channels there's no need to wait for them to start at all! The channels will queue up any interactions with them which occur before they start.

lahja/trio/endpoint.py
```diff
@@ -0,0 +1,905 @@
...
+
+    async def _monitor_subscription_changes(self) -> None:
+        while not self.is_stopped:
+            # We wait for the event to change and then immediately replace it
+            # with a new event.  This **must** occur before any additional
+            # `await` calls to ensure that any *new* changes to the
+            # subscriptions end up operating on the *new* event and will be
+            # picked up in the next iteration of the loop.
+            await self._subscriptions_changed.wait()
+            self._subscriptions_changed = trio.Event()
```
lithp: Now that we're in trio world this might be easier to implement with a channel. Using a channel would also allow for the batching of events that you wanted, when an event comes in this can first empty the queue and then send messages to the remotes.

tests/core/trio/test_trio_endpoint_wait_for.py
```diff
@@ -0,0 +1,34 @@
...
+        assert result.value == "test"
+        done.set()
+
+    async with trio.open_nursery() as nursery:
+        nursery.start_soon(_do_wait_for)
+
+        await bob.wait_until_endpoint_subscribed_to(alice.name, EventTest)
+
+        await bob.broadcast(event)
+        await done.wait()
```
lithp: I think this test could call `alice.wait_for` directly instead of waiting on another coro to do it.

lahja/trio/endpoint.py
pipermerriam: Ah, because it needs access to the nursery to run the endpoint in the background and I much prefer trying to keep access to the nursery isolated so this indirection is the cost.

lahja/trio/endpoint.py
pipermerriam: I renamed this to `_inbound_send_channel` (and there is a new `_outbound_send_channel` to signal the direction of message flow.

lahja/trio/endpoint.py
pipermerriam: I think you're right but it feels roughly equivalent and I'm inclined to leave this as-is.

tests/core/trio/test_trio_endpoint_wait_for.py
pipermerriam: No, it requires that the `alice.wait_for` be run first since the subscription needs to be in place for alice to receive it.

PR #130, opened 2019-06-12
consistent organization of code

pipermerriam:
## What was wrong?

There was inconsistency in how code was organized in both the base implementation and the asyncio one.

## How was it fixed?

Sorted out all code according to functionality and ensured that both the base and the asyncio implementation have everything organized *roughly* in the same order.

Also moved the compression logic and the endpoint management logic out of the `Asyncio` implementation as it can be shared by the trio implementation.

Also renamed some methods to be `_` prefixed to signal they are private.

**This should have zero meaningful functionality changes**

#### Cute Animal Picture

12-05-27 Sugar (12)C750


lahja/asyncio/endpoint.py
```diff
@@ -464,139 +532,14 @@ def connect_to_endpoints_nowait(self, *endpoints: ConnectionConfig) -> None:
...
-    @asynccontextmanager  # type: ignore
-    async def serve(cls, config: ConnectionConfig) -> AsyncIterator["AsyncioEndpoint"]:
-        endpoint = cls(config.name)
-        async with endpoint.run():
-            await endpoint.start_server(config.path)
-            yield endpoint
+        await asyncio.wait_for(
+            self.wait_until_connected_to(config.name),
+            timeout=constants.ENDPOINT_CONNECT_TIMEOUT,
+        )
```
pipermerriam: I added this timeout.

lahja/asyncio/endpoint.py
```diff
@@ -440,15 +511,12 @@ def connect_to_endpoints_nowait(self, *endpoints: ConnectionConfig) -> None:
 
     async def _await_connect_to_endpoint(self, endpoint: ConnectionConfig) -> None:
         await wait_for_path(endpoint.path)
-        # sleep for a moment to give the server time to be ready to accept
-        # connections
-        await asyncio.sleep(0.01)
```
pipermerriam: I removed this sleep.

pipermerriam:
:+1: I guess my intent was to say that under the assumption that `EndpointAPI` is our definition of *public API* this doesn't change anything.  I'll get release notes changes in place separately.

PR #131, opened 2019-06-12
Use context manager api for remote endpoint running

pipermerriam:
Builds on #130 

## What was wrong?

The mechanisms for starting a background process in `trio` vs `asyncio` are not compatible.  In `trio`, our `start` methods have to be passed a nursery object in order for them to be capable of starting things up and then returning.

The `BaseEndpointAPI` defined a `start` method which results in the base implementation not being compatible with what we need for the trio implementation to fit within the model.

## How was it fixed?

Converted it to use the same async context manager pattern as endpoints via `remote.run()`, delegating the actual startup and shutdown logic to the actual implementation.

#### Cute Animal Picture

10-05-30 Bella (181) C650

### DIFF
--- a/lahja/base.py
+++ b/lahja/base.py
@@ -97,12 +97,29 @@ class RemoteEndpointAPI(ABC):
     def is_stopped(self) -> bool:
         ...
 
+    #
+    # Running API
+    #
     @abstractmethod
-    async def start(self) -> None:
+    def run(self) -> AsyncContextManager["RemoteEndpointAPI"]:
+        """
+        Context manager API for running endpoints.
+
+        .. code-block::
+
+            async with endpoint.run() as endpoint:
+                ... # endpoint running within context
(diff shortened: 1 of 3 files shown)


lahja/base.py
```diff
@@ -366,6 +383,13 @@ def serve(cls, config: ConnectionConfig) -> AsyncContextManager["BaseEndpoint"]:
         """
         ...
 
+    @abstractmethod
+    async def connect_to_endpoints(self, *endpoints: ConnectionConfig) -> None:
```
cburgdorf: Unrelated to this PR but should we just remove the non-variadic `connect_to_endpoint`. Since the variadic one can be called with `1..n` connection configs it seems unnecessary to keep an extra API for the single parameter use case, no?

cburgdorf: https://github.com/ethereum/lahja/pull/137

PR #134, opened 2019-06-13
Drop un-needed ignore statements on the async context managers

pipermerriam:
## What was wrong?

These ignore statements "may" no longer be needed.

## How was it fixed?

Remeoved them.

#### Cute Animal Picture

monifa-hippo-7


pipermerriam:
weird, this passes in my other branch

PR #136, opened 2019-06-13
Cleanup @property stuff on BaseRemoteEndpoint

pipermerriam:
## What was wrong?

There were some errors in the implementation that needed cleaning up

## How was it fixed?

Fixed them (need to add tests)

#### Cute Animal Picture

6a010535647bf3970b0105367c839d970b-800wi

### DIFF
--- a/lahja/base.py
+++ b/lahja/base.py
@@ -221,12 +221,15 @@ class BaseRemoteEndpoint(RemoteEndpointAPI):
     async def wait_stopped(self) -> None:
         await self._stopped.wait()
 
+    @property
     def is_running(self) -> bool:
-        return not self.is_stopped and self.running.is_set()
+        return not self.is_stopped and self._running.is_set()
 
+    @property
     def is_ready(self) -> bool:
         return self.is_running and self._ready.is_set()
 
+    @property
     def is_stopped(self) -> bool:
         return self._stopped.is_set()
 


lahja/base.py
```diff
@@ -221,12 +221,15 @@ def name(self) -> str:
     async def wait_stopped(self) -> None:
         await self._stopped.wait()
 
+    @property
     def is_running(self) -> bool:
-        return not self.is_stopped and self.running.is_set()
+        return not self.is_stopped and self._running.is_set()
```
cburgdorf: I'm really confused by this. How is it possible that neither `mypy` nor `flake8` caught this?

pipermerriam: In my local env. just pushed something to see if CI catches it.

pipermerriam: it did https://circleci.com/gh/pipermerriam/lahja/2245

cburgdorf: Yes, it seems it catches it *after* this change but not on current master which is really weird. Also, the `@property` decorators also need to be applied on the `RemoteEndpointAPI`.

cburgdorf: In fact, as soon as you remove the `@property` from `is_stopped` it won't catch this error anymore. This feels like a weird mypy bug to me.

pipermerriam: That's very odd.

cburgdorf: Ok, here is a minimal test case.

```python
class TestCase:

    def foo(self) -> bool:
        return False and self.meh()
```

This is not caught by `mypy`. Do you think this is expected behavior?

cburgdorf: I opened an issue against mypy https://github.com/python/mypy/issues/6990

PR #138, opened 2019-06-13
Add missing decorators

cburgdorf:
## What was wrong?

@property decorators are missing in the base class

## How was it fixed?

Added them

#### Cute Animal Picture


cburgdorf:
@pipermerriam I guess this went unnoticed because I mentioned it after I already approved the PR :)

Issue #63, opened 2019-05-17
Metaclass API for defining events


pipermerriam:
### What is wrong

Defining new events requires un-necessary boilerplate.  Events are typically purely for data transfer and rarely have extra methods or properties.

### How can it be fixed?

It'd be really nice to be able to do this:

```python
class NewUser(DataEvent):
    username: str
    created_at: datetime.datetime
    age: int

event = NewUser('piper', datatime.datetime.now(), 99)
```

Doing this with metaclasses is quite easy and it would reduce the boilerplate in defining new events significantly.

cburgdorf:
We *could* also just force it on the user to add a `@dataclass` decorator on the events. I opened an PR to demonstrate this with the examples: https://github.com/ethereum/lahja/pull/139

pipermerriam:
dataclasses could be a nice way to do this but I'm not inclined to force it.

cburgdorf:
>but I'm not inclined to force it.

It would be totally up to the consumer to decide to use them or not. Here's a PR to show case that for Trinity https://github.com/ethereum/trinity/pull/726 I only converted a few places there but I happily convert all of them if you are :+1: on that change.

PR #139, opened 2019-06-14
Use dataclass to reduce boilerplate in event definitions

cburgdorf:
## What was wrong?

This is an alternative to #63 relying on the new `@dataclass` decorator that was added with Python 3.7. There's a backport for Python 3.6 available as well: https://pypi.org/project/dataclasses/

## How was it fixed?

Use dataclasses in examples

#### Cute Animal Picture


cburgdorf:
It fails for the `py36-examples` job because I haven't used the Python 3.6 fallback here but we *could* do that. But on the other hand, if it's easy to build our own `DataEvent` with a metaclass that might be even better.

cburgdorf:
Ok, I went ahead and added the `dataclasses` backport for the `py36` job. I read up a bit on https://docs.python.org/3/library/dataclasses.html and I'm leaning towards just going with that as dataclasses come with a bunch of other features and so this may just end up as the more flexible solution. Thoughts?

Issue #140, opened 2019-06-15
Declarative cross implementation test suite tools


pipermerriam:
## What was wrong?

- lots of duplication between the `trio` and `asyncio` based test suites.
- lots of things that behave differently when they occur across different processes.

## How can it be fixed?

I've been imagining this concept that I'll call an "Endpoint Driver".  It would be a declarative, implementation agnostic mechanism for defining a sequential set of actions that should be taken on an endpoint, with one of those actions being something like just running python code.

This tool should allow us to write implementation agnostic tests allowing us to define our test suite once and run it across any number of different combinations like:

- both endpoints of the same implementation in the same process
- different implementations in the same process
- same/different implementations in different processes.

I *suspect* the tests will be much more readable too.

pipermerriam:
```python
server_driver = compose_driver(
    connect_until_connected_to(client_config),
    wait_for(SomeEvent).add_handler(lambda ev: assert isinstance(ev, SomeEvent)),
)
client_driver = compose_driver(
    connect_to_endpoint(server_config),
    broadcast(SomeEvent()),
)
```

Then there's probably a *backend* type component for how to run each endpoint type as well as something to manage doing things in their own processes.

pipermerriam:
Done with #141 opening new issues to convert the rest of the test suite to use this...

PR #141, opened 2019-06-19
Endpoint Driver

pipermerriam:
## What was wrong?

Our test suite has a large amount of duplication across the trio and asyncio tests.

## How was it fixed?

Wrote a suite of tools for running declarative tests against endpoints that can run in the same test suite.

- Drivers: 
    - These are instructions on how to run and endpoint and perform actions with it.
- Engines
    - These are how we can run endpoints agnostic to whether they are `trio` or `asyncio` 
- Runners
    - These are how we can run the same test code across different endpoint configurations like both in the same process, both in different processes, or even different endpoint implementations in different processes. 

Each driver is composed of an *initializer* and a sequence of actions.  There is a built-in timeout enforced on actions so that the test suite doesn't hanging.

Each action also emits log output which is really helpful for debugging.

#### Cute Animal Picture


pipermerriam:
related to https://github.com/ethereum/lahja/issues/140

pipermerriam:
@cburgdorf or @lithp I'd be curious to get feedback on this now that I've polished it a little.

lahja/tools/drivers/driver.py
```diff
@@ -0,0 +1,328 @@
...
+    result = await endpoint.wait_for(event_type)
+    logger.debug('[%s].wait_for(%s) RECEIVED', endpoint, event_type)
+    if on_event is not None:
+        if asyncio.iscoroutinefunction(on_event):
+            await on_event(endpoint, result)
+        else:
+            on_event(endpoint, result)
+
+
+def wait_for(event_type: Type[BaseEvent], on_event: EventHandlerFn = None) -> AsyncAction:
```
cburgdorf: Looks like this needs to be `Optional[EventHandlerFn]`

lahja/tools/drivers/driver.py
```diff
@@ -0,0 +1,328 @@
...
+    logger.debug('[%s].request(%s) GOT RESPONSE: %s', endpoint, event, response)
+    if on_response is not None:
+        if asyncio.iscoroutinefunction(on_response):
+            await on_response(endpoint, response)
+        else:
+            on_response(endpoint, response)
+
+
+def request(event: BaseRequestResponseEvent[BaseEvent],
+            config: BroadcastConfig = None,
```
cburgdorf: Same here, some missing `Optional` (But I guess our strict type checking in this project will catch this anyway)

lahja/tools/drivers/driver.py
```diff
@@ -0,0 +1,328 @@
...
+    return Initializer(_run_endpoint, name=name)
+
+
+#
+# Serve request
+#
+GetResponseFn = Callable[[EndpointAPI, BaseRequestResponseEvent], BaseEvent]
+
+
+async def _serve_request(engine: EngineAPI,
```
cburgdorf: Looks like `engine` can be dropped from this and some of the following functions.

tests/core/common/conftest.py
```diff
@@ -0,0 +1,24 @@
...
+    if request.param == 'trio-runner':
+        return TrioRunner()
+    elif request.param == 'asyncio-runner':
+        return AsyncioRunner()
+    elif request.param == 'trio-isolated':
+        return IsolatedHomogenousRunner(TrioRunner())
+    elif request.param == 'asyncio-isolated':
+        return IsolatedHomogenousRunner(TrioRunner())
+    elif request.param == 'mixed-isolated':
+        return IsolatedHeterogenousRunner(TrioRunner(), AsyncioRunner())
```
cburgdorf: All the additional coverage we get through this is just brilliant :+1:

tests/core/common/test_endpoint_broadcast.py
```diff
@@ -0,0 +1,40 @@
+from lahja import BaseEvent, ConnectionConfig
+from lahja.tools import drivers as d
```
cburgdorf: meh single letter alias

cburgdorf: Actually, looking at the saved space we gain, it's not that bade to make an exception here!

tests/core/common/test_endpoint_request_response.py
```diff
@@ -0,0 +1,35 @@
...
+
+
+def test_request_response(runner, ipc_base_path):
+    server_config = ConnectionConfig.from_name("server", base_path=ipc_base_path)
+
+    received = multiprocessing.Event()
+
+    server = d.driver(
+        d.serve_endpoint(server_config),
+        d.serve_request(Request, get_response=lambda endpoint, event: Response()),
```
cburgdorf: Piper :heart: `lambda`

lahja/tools/drivers/driver.py
```diff
@@ -0,0 +1,328 @@
...
+    logger.debug('[%s] waiting for request: %s', endpoint, request_type)
+    request = await endpoint.wait_for(request_type)
+    logger.debug('[%s] received request: %s', endpoint, request)
+    response = get_response(endpoint, request)
+    await endpoint.broadcast(response, config=request.broadcast_config())
+    logger.debug('[%s] sent response: %s', endpoint, response)
+
+
+def serve_request(request_type: BaseRequestResponseEvent[BaseEvent],
+                  get_response: GetResponseFn) -> AsyncAction:
```
cburgdorf: Would `repond_with` feel a little more ergonomic than `get_response`?

lahja/tools/drivers/driver.py
pipermerriam: It needs to be present because all of the driver `Action` APIs have to have the same signature.  It's not used in most of them, but it's present to give access to the engine for a case where a `sleep` needs to occur.

pipermerriam:
@cburgdorf thanks for the review.  As for path to merging this, what do you think about me cleaning it up, writing some docs for it and we can merge as-is and then do the remaining cleanup of the tests (and creation of other necessary driver functions) in subsequent PRs to keep things smaller?

cburgdorf:
Sounds good to me!

pipermerriam:
Ready for a real review then.  Got tests passing and did some basic cleanup as well as writing up very rough/basic documentation.

PR #144, opened 2019-07-25
graceful failure if IPC file is missing

pipermerriam:
## What was wrong?

If for some reason the IPC file is missing during server shutdown, the `asyncio` endpoint throws a `FileNotFoundError` while trying to remove it.

## How was it fixed?

Catch this error and silence it when shutting down the server.

#### Cute Animal Picture

llamas2


lahja/asyncio/endpoint.py
```diff
@@ -462,7 +462,10 @@ def _stop_server(self) -> None:
         for task in self._server_tasks:
             task.cancel()
 
-        self.ipc_path.unlink()
+        try:
+            self.ipc_path.unlink()
+        except FileNotFoundError:
```
cburgdorf: Might still be worth to issue a warning or debug message? Just an idea, feel free to ignore.

cburgdorf:
I'm going to merge this one because I want to see if it triggeres a new RTD build (turns out they were broken for two months and I just attempted to fix it)

cburgdorf:
@pipermerriam ok, it's still not triggering new RTD builds. I'm not sure but maybe I'm missing some rights to make configurations on GitHub? I *think* you also magically fixed that for Trinity. Can you take a look?

pipermerriam:
Should be fixed now

PR #145, opened 2019-07-25
Proper cancellation of asyncio tasks

pipermerriam:
## What was wrong?

When you cancel an `asyncio` task you are supposed to await it to give it a chance to have the cancellation error thrown into it and for it to cleanup.  We were not doing this.

## How was it fixed?

Added a `try/except` to each place we do this type of cancellation with the proper `await` statement.

#### Cute Animal Picture

16-02-01 MacTravish (12)C1000

### DIFF
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -171,6 +171,10 @@ class AsyncioRemoteEndpoint(BaseRemoteEndpoint):
             return
         self._stopped.set()
         self._task.cancel()
+        try:
+            await self._task
+        except asyncio.CancelledError:
+            pass
 
 
 TFunc = TypeVar("TFunc", bound=Callable[..., Any])
@@ -294,7 +298,7 @@ class AsyncioEndpoint(BaseEndpoint):
         try:
             yield self
         finally:
-            self._stop()
+            await self._stop()
 
     @check_event_loop
     async def _start(self) -> None:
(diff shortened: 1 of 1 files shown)


lahja/asyncio/endpoint.py
```diff
@@ -313,19 +317,23 @@ def run(self, *args, **kwargs):  # type: ignore
...
         if not self.is_running:
             return
 
-        self._stop_server()
+        await self._stop_server()
 
         self._running = False
 
-        for task in self._endpoint_tasks:
+        for task in tuple(self._endpoint_tasks):
```
cburgdorf: Curious why we need to transform the `set` into a `tuple` here?

lahja/asyncio/endpoint.py
```diff
@@ -452,15 +460,19 @@ def _stop(self) -> None:
...
-    def _stop_server(self) -> None:
+    async def _stop_server(self) -> None:
         if not self.is_serving:
             return
         self._serving = False
 
         self._server.close()
 
-        for task in self._server_tasks:
+        for task in tuple(self._server_tasks):
```
cburgdorf: and here

lahja/asyncio/endpoint.py
pipermerriam: because by doing the cancellation properly the task actually finishes, executing it's cleanup task which removes it from the list of tasks and then python complains that the size changed during iteration so we have to freeze it before iterating over it.

cburgdorf:
That makes sense!

PR #154, opened 2019-08-27
Increase ENDPOINT_CONNECT_TIMEOUT to 30 seconds

cburgdorf:
### What was wrong?

We recently introduced a connection timeout. This turned out problematic in Trinity as 2s proved as too short.

### How was it fixed?

Increase timeout to 30 seconds. I think that we should pick this value *very* conservatively. For a generic library such as lahja, it is very hard to decide these things. For some applications a endpoint not starting to respond after 2 seconds may be a sign of trouble whereas for others it may be perfectly expected behavior. E.g. one could imagine waiting for an endpoint to come up eternally because the reason whether it should come up or not may depend on some user action.

So, *ideally* I'd actually make this value entirely configurable and allow `None` as a valid timeout to wait until forever but it looks like the `trio.fail_after` API does not allow waiting until forever (`asyncio.wait_for` does!).

So for now, setting this to `30` strikes me as a simple solution to at least resolve the issue for Trinity.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture


cburgdorf:
This would resolve https://github.com/ethereum/trinity/pull/917

cburgdorf:
I'm on board :+1:

cburgdorf:
Ok, let's take it as-is for now to unblock the Trinity upgrade :+1:

PR #157, opened 2019-09-06
Ensure lahja closes asyncio streams

pipermerriam:
### What was wrong?

We weren't closing the `asyncio` streams when and endpoint shuts down.

In `trinity` this caused a *lot* of `ResourceWarning:` warnings to end up in the terminal on shutdown.

### How was it fixed?

Added `ConnectionAPI.close()` and ensure that it's called during the closing of the `RemoteConnection`.

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

cat-angry-posing-funny-cat-pics


lahja/asyncio/endpoint.py
```diff
@@ -106,7 +111,7 @@ def __init__(self, reader: StreamReader, writer: StreamWriter) -> None:
             # We don't do a pre-check for a closed writer since this is a performance critical
             # path. We also don't want to swallow runtime errors unrelated to closed handlers.
             if "handler is closed" in str(err):
-                self.logger.warning("Failed to send %s. Handler closed.", message)
+                self.logger.debug("Failed to send %s. Handler closed.", message)
```
pipermerriam: Changed to `debug` because otherwise this dumps warnings on shutdown.

lahja/tools/drivers/actions.py
```diff
@@ -87,26 +87,22 @@ def wait_for(
...
-
 def serve_request(
     request_type: Type[BaseRequestResponseEvent[BaseEvent]], get_response: GetResponseFn
 ) -> AsyncAction:
     """
     Wait for an event of the provided ``request_type`` and respond using the
     response event returned by the provide ``get_response`` function.
     """
-    return wait_for(
-        request_type, functools.partial(_serve_response, get_response=get_response)
```
pipermerriam: This was producing a value that `asyncio.iscoroutinefunction` wasn't detecting as being an awaitable.

pipermerriam:
not definitive but I also think is a major culprit for how long trinity takes to shutdown.  shutdowns seem much faster with this change.

PR #165, opened 2019-12-02
WIP: Synchronously unlink IPC file in TrioEndpoint

gsalgado:
### What was wrong?

It seems wrong to me to use an async API when cleaning up the IPC file as that will simply not run when the current trio context is cancelled. As a concrete example, that is what happened in https://github.com/ethereum/trinity/pull/1325, which was always leaving behind the IPC file for the discovery service

### How was it fixed?


gsalgado:
Had to update to mypy 0.740 because of https://circleci.com/gh/ethereum/lahja/2754?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link but then now we're getting that other failure

lahja/trio/endpoint.py
```diff
@@ -599,12 +597,12 @@ def is_server_stopped(self) -> bool:
         del self._server_nursery
 
         try:
-            await self.ipc_path.unlink()
+            self.ipc_path.unlink()
```
gsalgado: Alternatively, we could continue using the async API (via trio.Path) but change the try/except to catch trio.Cancelled, then synchronously unlink the file and re-raise trio.Cancelled. But I really don't see the point of using the async API here in the first place

tests/core/driver/test_endpoint_driver.py
```diff
@@ -4,7 +4,7 @@
 import pytest
 
 from lahja import BaseEvent, ConnectionConfig
-from lahja.tools.driver import (
+from lahja.tools.drivers import (
```
gsalgado: There's no `lahja.tools.driver` package/module. I've no idea how CI didn't complain about this before?

cburgdorf: It looks like we forgot to actually run this test

https://github.com/ethereum/lahja/blob/b34700f3bdb040b91a943544f914109c2b0b883d/tox.ini#L27-L29

lahja/trio/endpoint.py
cburgdorf: I'm not familar with the trio APIs tbh. From what I'm seeing here I agree that synchronously unlinking should be fine. But maybe @pipermerriam who wrote the trio implementation has more insight.

cburgdorf:
If the mypy error is about a trio API that is missing in trio maintained stub files, then an `type: ignore` + comment seems like an appropriate solution for now. Also opening an issue upstream would be nice then.

lahja/trio/endpoint.py
pipermerriam: I don't think the async filesystem APIs are actually that good in trio right now so I think this is just fine.  Last I remember the main trio maintainer said they were implemented using threads and were kind of gross.  :+1:

gsalgado:
The mypy error seems legit as it's only MemorySendChannel that has a `send_nowait()`

pipermerriam:
> The mypy error seems legit as it's only `MemorySendChannel` that has a `send_nowait()`

I ran into this in the trinity codebase.  AFAIK there is no good way around it.  I'd use a `type: ignore`, comment as to why, I'll open an issue in the `trio` codebase to see what they can do to fix this.

pipermerriam:
https://github.com/python-trio/trio/issues/1327

tests/core/driver/test_endpoint_driver.py
gsalgado: It looks like we don't run the tests/core/common tests either. I've changed tox.ini in an attempt to correct that but they're still not being executed on CI. Any idea what's wrong?

cburgdorf: Good catch! 

>I've changed tox.ini in an attempt to correct that but they're still not being executed on CI. Any idea what's wrong?

They also have to be configured in `circleci/config.yml`

https://github.com/ethereum/lahja/blob/b34700f3bdb040b91a943544f914109c2b0b883d/.circleci/config.yml#L78-L89

cburgdorf:
@gsalgado FYI the remaining `lint` error is because this repo uses `black` formatting so you need to rerun `black <dir>` to have it reformat the file for you.

gsalgado:
@cburgdorf any suggestions on how to fix https://circleci.com/gh/ethereum/lahja/2897?utm_campaign=vcs-integration-link&utm_medium=referral&utm_source=github-build-link?

PR #166, opened 2019-12-03
Bump version: 0.15.0 → 0.15.1

gsalgado:


gsalgado:
@cburgdorf py36-core-common is failing intermittently. any chance you can have a look and see if it's anything obvious? otherwise I'm tempted to file a bug and disable it for now as it is one that was enabled in my last merge

PR #171, opened 2019-12-06
Add Matomo tracking

cburgdorf:
### What was wrong?

Currently we do not have a good insight into our target group. In order to improve both Trinity itself and the documentation, it makes sense to aggregate some data.

### How was it fixed?

Add Matomo Tracking to Docs site.

Matomo is an Open Source web analytics platform that allows us
to get better insights and optimize for our audience without
the negative consequences of other compareable platforms.

Read more: https://matomo.org/why-matomo/

### To-Do

- [x] Clean up commit history

#### Cute Animal Picture

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -145,7 +145,8 @@ html_theme = 'sphinx_rtd_theme'
 html_static_path = ['_static']
 
 def setup(app):
-    app.add_stylesheet("css/custom.css")
+    app.add_css_file("css/custom.css")
+    app.add_js_file("js/matomo.js")
 
 # Add any extra paths that contain custom files (such as robots.txt or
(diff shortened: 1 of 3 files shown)


docs/conf.py
```diff
@@ -145,7 +145,8 @@
 html_static_path = ['_static']
 
 def setup(app):
-    app.add_stylesheet("css/custom.css")
+    app.add_css_file("css/custom.css")
```
cburgdorf: The old name of `add_stylesheet` is deprecated. We have pinned sphinx to an older version in Trinity hence it's using the old name there. Just in case someone is wondering about this change.

pipermerriam:
rage merge so I can see if this is working

pipermerriam:
huzzah, it works

PR #176, opened 2020-03-31
Raise error if broadcast/request has no subscribers

cburgdorf:
### What was wrong?

It is easy to end up in a scenario where important events are forgotten to get subscribed to. This can lead to fatal bugs such as [Trinity not propagating any transactions to its peers](https://github.com/ethereum/trinity/pull/1649)

### How was it fixed?

1. By default any broadcasted event as well as any send request will throw a `NoSubscriber` exception if there are no subscribers to said event/request at the time when the event/request is being made.

2. Events/Requests can still be fired if there are no subscribers but it needs to be made explicit via `BroadcastConfig(raise_if_no_subscribers=False)`. As an example, this can be useful for events that are meant for extensibility to be subscribed to conditionally based on the fact whether a specific plugin is installed or not.

#### Alternative designs

I thought about using the `wait_until_any_endpoint_subscribed_to` API to allow *waiting* for subscriptions to be made before the event is send out but this has drawbacks:

1. It wouldn't work for the synchronous `broadcast_nowait` API

2. It raises the question how long should be waited for subscriptions. This could be made configurable but it feels like a weaker API.

3. When an endpoint connects to a remote endpoint, it does already wait for the initial batch of subscriptions for the remote endpoint anyway

4. Even if a subscription would be expected to arrive late it would be easy enough for the developer to use the `wait_until_any_endpoint_subscribed_to` manually *once* before firing the event.

### To-Do

- [ ] Clean up commit history

- [ ] Add entry to the [release notes](https://github.com/ethereum/lahja/blob/master/newsfragments/README.md)

#### Cute Animal Picture

### DIFF
--- a/lahja/base.py
+++ b/lahja/base.py
@@ -37,7 +37,7 @@ from .common import (
     SubscriptionsAck,
     SubscriptionsUpdated,
 )
-from .exceptions import ConnectionAttemptRejected, RemoteDisconnected
+from .exceptions import ConnectionAttemptRejected, NoSubscribers, RemoteDisconnected
 from .typing import ConditionAPI, EventAPI, LockAPI, RequestID
 
 TResponse = TypeVar("TResponse", bound=BaseEvent)
@@ -654,6 +654,24 @@ class BaseEndpoint(EndpointAPI):
             (remote.name, remote.get_subscribed_events())
             for remote in self._connections
         )
+
+    def maybe_raise_no_subscribers_exception(
+        self, config: Optional[BroadcastConfig], event_type: Type[BaseEvent]
+    ) -> None:
+        """
+        Check the given ``config`` and ``event_type`` and raise a
+        :class:`~lahja.exceptions.NoSubscriber` if no subscribers exist for the ``event_type``
+        when they at least one subscriber is expected.
+        """
+
+        if config is not None:
+            if config.require_subscriber:
+                return
+            if config.filter_event_id is not None:
+                # This is a response to a request
+                return
+        elif not self.is_any_endpoint_subscribed_to(event_type):
+            raise NoSubscribers(f"No subscribers for: {event_type}")
 
     async def wait_until_connections_change(self) -> None:
         """
--- a/lahja/common.py
+++ b/lahja/common.py
@@ -36,11 +36,13 @@ class BroadcastConfig:
         filter_endpoint: Optional[str] = None,
         filter_event_id: Optional[RequestID] = None,
         internal: bool = False,
+        require_subscriber: bool = True,
     ) -> None:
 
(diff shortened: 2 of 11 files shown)


lahja/common.py
```diff
@@ -36,11 +36,13 @@ def __init__(
         filter_endpoint: Optional[str] = None,
         filter_event_id: Optional[RequestID] = None,
         internal: bool = False,
+        raise_if_no_subscribers: bool = True,
     ) -> None:
 
         self.filter_endpoint = filter_endpoint
         self.filter_event_id = filter_event_id
         self.internal = internal
+        self.raise_if_no_subscribers = raise_if_no_subscribers
```
cburgdorf: @pipermerriam I'm almost certain you'll have an idea for a better name for this :sweat_smile:

lahja/base.py
```diff
@@ -655,6 +655,19 @@ def get_connected_endpoints_and_subscriptions(
             for remote in self._connections
         )
 
+    def maybe_raise_no_subscribers_exception(
```
pipermerriam: Seems like this should be namespaced as a private API.

lahja/common.py
pipermerriam: A good policy is to look for the name that doesn't have any contradictions.  The `no_subscribers` is the part that I think sparks the most confusion here.

maybe:

- `require_active_subscription`
- `require_subscriber`
- `is_subscription_required`
- `is_subscriber_required`

Any of those do it for you?

cburgdorf: Ha, `require_subscriber` is so obvious now that you bring it up :see_no_evil:

lahja/base.py
cburgdorf: It can not be *private*  because because we want to call it from inherited classes. And if we mark it private it will fall out of the docs (rightfully so). In Java or .NET terms, this would be *protected*  or *internal* if we want to express it in Solidity terms. I've had this a few times before that I really wished we had a convention on how to prefix *protected* members. Anything that isn't an underscore to keep it in the docs. But a prefix like `protected` or `internal` may make an odd combination with the actual function name and also eats quite a bit of horizontal space...

cburgdorf:
>as you bring this into trinity, pay attention for whether we should look into adding another API for automatically waiting for the subscription prior to broadcasting (which would effectively wait indefinitely and require timeout management to occur outside the call).

It looks like we don't need it. When endpoints connect to their remote endpoints they await the initial batch of subscriptions anyway and so all the changes to [update Trinity](https://github.com/ethereum/trinity/pull/1653) are pretty straight forward.

cburgdorf:
@pipermerriam please let me know if you are fine with me cutting a new release today

pipermerriam:
:shipit:

PR #179, opened 2020-07-31
Ensure asyncio Endpoint.stream implementation yields to event loop

pipermerriam:
### What was wrong?

When using code like this, the async generator does not natively yield to the event loop:

```python
async for event in endpoint.stream(...):
    # do something
```

### How was it fixed?

Added a call to `await asyncio.sleep(0)` in the implementation so that it would not block other coroutines from executing.

#### Cute Animal Picture

hawaiianchicken

### DIFF
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -1,5 +1,5 @@
 import asyncio
-from asyncio import StreamReader, StreamWriter
+from asyncio import QueueEmpty, StreamReader, StreamWriter
 from collections import defaultdict
 import functools
 import itertools
@@ -756,10 +756,14 @@ class AsyncioEndpoint(BaseEndpoint):
         try:
             for _ in iterations:
                 try:
-                    if queue.empty():
+                    try:
+                        yield queue.get_nowait()
+                    except QueueEmpty:
                         yield await queue.get()
                     else:
-                        yield queue.get_nowait()
+                        # We sleep here to yield to the event loop in the case
+                        # that we were able to get an item from the queue
+                        # without an actual `await`
                         await asyncio.sleep(0)
                 except GeneratorExit:
                     break
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -756,11 +756,13 @@ class AsyncioEndpoint(BaseEndpoint):
         try:
             for _ in iterations:
                 try:
-                    yield await queue.get()
+                    if queue.empty():
+                        yield await queue.get()
+                    else:
+                        yield queue.get_nowait()
+                        await asyncio.sleep(0)
                 except GeneratorExit:
                     break
-                else:
-                    await asyncio.sleep(0)
         finally:
             self._queues[event_type].remove(casted_queue)
             if not self._queues[event_type]:
--- a/lahja/asyncio/endpoint.py
+++ b/lahja/asyncio/endpoint.py
@@ -759,6 +759,8 @@ class AsyncioEndpoint(BaseEndpoint):
                     yield await queue.get()
                 except GeneratorExit:
                     break
+                else:
+                    await asyncio.sleep(0)
         finally:
             self._queues[event_type].remove(casted_queue)
             if not self._queues[event_type]:
--- a/newsfragments/179.bugfix.rst
+++ b/newsfragments/179.bugfix.rst
@@ -1,0 +1,1 @@
+Ensure that the ``asyncio`` implementation releases control of the event loop on each iteration of the async iterator produced by `EndpointAPI.stream`


lahja/asyncio/endpoint.py
```diff
@@ -759,6 +759,8 @@ async def stream(
                     yield await queue.get()
                 except GeneratorExit:
                     break
+                else:
+                    await asyncio.sleep(0)
```
gsalgado, line 763: The `await queue.get()` above should always yield back to the event loop

carver: So I decided it was time to finally learn some more about asyncio. So now I'll be a little pedantic, because I've misunderstood this idea that "await yields to the event loop." It can, but there are plenty of times it won't. So a slightly more precise version in this case is:

> The `await queue.get()` above ~should always~ *will* yield back to the event loop, *but only when the queue is empty*


This modified example from [this stackoverflow post](https://stackoverflow.com/a/59780868/8412986) helped me understand it a bit more:
```py
import asyncio
from itertools import count

counter = count()


def task_print(message):
    print(message)


async def async_print(message):
    task_print(message)


async def next_val():
    return next(counter)


async def arange(size):
    for _ in range(size):
        yield await next_val()


async def original_task():
    asyncio.create_task(async_print("This doesn't run until the very end because the generated None following the creation of this task is consumed by the `for` loop"))
    async for y in arange(3):
        await async_print(f"'ordinary' `yield` points (those which are consumed by the `current_task` itself) behave as ordinary without relinquishing control at the async/task-level; `y={y}`")
    task_print("Done with original task")

asyncio.get_event_loop().run_until_complete(original_task())
```
Prints:

> 'ordinary' `yield` points (those which are consumed by the `current_task` itself) behave as ordinary without relinquishing control at the async/task-level; `y=0`
> 'ordinary' `yield` points (those which are consumed by the `current_task` itself) behave as ordinary without relinquishing control at the async/task-level; `y=1`
> 'ordinary' `yield` points (those which are consumed by the `current_task` itself) behave as ordinary without relinquishing control at the async/task-level; `y=2`
> Done with original task
> This doesn't run until the very end because the generated None following the creation of this task is consumed by the `for` loop

The number of awaits doesn't matter, and it won't release at the beginning or middle or end of an await, even multiple layers down. Only when we await something that then `yield`s to the event loop.

pipermerriam: Less thorough than @carver but I did a test before and was surprised to find it didn't yield (when there were items in the queue).

gsalgado: Yeah, it makes sense that `await queue.get()` will only relinquish control to the loop (trying to avoid the term yield as we're talking about generators) when the queue is empty, so I guess the question we need to answer now is whether or not we want `stream()` to relinquish control to the loop after every item yielded? My gut feeling is that it'd make more sense to not worry about that here (but document its behavior clearly) and let consumers decide how often they want to relinquish control. The main reason being that I suspect in most cases the number of events streamed is relatively small so that we can stream them all in one go. And if a consumer is doing expensive operations for each of those events, it should be doing so in a cooperative fashion, which would relinquish control back to the event loop. It's also important to notice that when the queue is empty it will relinquish control one extra time, unnecessarily.

gsalgado: I couldn't find any async iterators/generators in the stdlib so that we could at least be consistent with, but the websocket library has one which will yield all queued messages before relinquishing control to the loop: https://github.com/aaugustin/websockets/blob/master/src/websockets/protocol.py#L439-L523

pipermerriam: I just checked and `trio.ReceiveChannel` used as an async iterator yields to other tasks on each item.  I'd advocate following their lead (which is what this PR does).

gsalgado: Fair enough

carver: So factors at play:
- added cost of yielding to the event loop in between each event (I'm not sure what the order of magnitude cost is here)
- risk of holding the event loop too long
- when several events come through simultaneously, the later messages will have higher latency because of yielding control

It's not an obvious answer here, especially because of the unknown cost of yielding control. My instinct says that we should yield at every element here. It's really difficult to detect and track down where the event loop is holding control for too long (at least in my install, for some reason). It can have pretty dramatic effects on performance.

One small suggestion: we can avoid yielding to the event loop twice by checking if the queue is empty before the `.get()`. If it was, then we can skip the `sleep(0)`.

Talking to myself: it would be kind of neat to be able to call `loop.time_since_yield()` to know how long the current coro has been holding control of the loop.

carver (approved):

lahja/asyncio/endpoint.py
pipermerriam: I'll update to do the more intelligent conditional `sleep(0)` based on whether we've already yielded due to the queue being empty.

pipermerriam:
Either of you want to give this one more quick sanity check?

lahja/asyncio/endpoint.py
```diff
@@ -756,7 +756,11 @@ async def stream(
         try:
             for _ in iterations:
                 try:
-                    yield await queue.get()
+                    if queue.empty():
```
gsalgado, line 759: Maybe a comment explaining why we do this?

gsalgado (approved):
LGTM, just one small suggestion

